PR #1, opened 2025-11-25
Readd some missing lines.

Xenthio:
Building already adds these but that's inconvenient as it adds a bunch of staging changes for everyone if they're not already like this on the origin.

### DIFF
--- a/game/benchmark.runtimeconfig.json
+++ b/game/benchmark.runtimeconfig.json
@@ -6,6 +6,7 @@
       "version": "10.0.0"
     },
     "configProperties": {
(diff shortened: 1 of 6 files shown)


rodlopt:
GOGpolLW8AAAazX

kaikenny:
Screenshot 2025-11-17 174759

DrakeFruit:
very productive conversation going on here I see

Grocel:
First Blood🔥

handsomematt:
We'll probably add these to gitignore on our end

PR #4, opened 2025-11-25
Envmap Probe Time Slicing feature

K3rhos:
I had previously requested this feature here: [Envmap Probe Time Slicing](https://github.com/Facepunch/sbox-public/issues/691) but no one really paid attention to it.

So here it is, just like what Unity is doing, instead of rendering all the 6 faces of the cubemap in a single iteration, time slicing spread the rendering of the 6 different textures of a cubemap over 6 frames instead ! (1 frame = 1 face)

Capture d&#39;écran 2025-11-26 001249

This drastically improve realtime envmap probe performance while almost noticing no difference in visual quality !

Here is a quick performance comparison in the default Sweeper scene, with an EnvmapProbe set to Every Frame refresh with and without Time slicing feature enabled:

**Time Slicing Off**
Capture d&#39;écran 2025-11-26 001754

**Time Slicing On**
Capture d&#39;écran 2025-11-26 001734

**Performance Graph**
image

If there is anything you would suggest for me to improve before merging this PR, please let me know.


sampavlovic:
Rebase it without menu

K3rhos:
> Rebase it without menu

Okay I think this is good now

PR #5, opened 2025-11-26
Fix file locking and retry logic in DownloadPublicArtifacts with improved UX 

Xenthio:
Downloads previously failed intermittently with "file used by another process" errors. This adds a retry to file downloads, aswell as improves UX to not spam the console.
image


Xenthio:
gonna reopen less ass

PR #9, opened 2025-11-26
Hammer Entity Selection Fix

kaikenny:
As a hammer user, this as been a really annoying bug and has been here for like a month at least now, so I decided to fix it.

Not much changed, Hammer still works as it should, not any bugs have been created that weren't already there (I think), etc. 
This is not too much of a request hopefully, this is my first pull request so hope I didn't do anything wrong, thanks!

https://github.com/user-attachments/assets/44b7f44d-3c0e-4120-b9bb-d28f6b9de76a

Before: No matter what was selected only info_player_start would always place
Now: Correctly works, and places what's selected.


kaikenny:
im tired so do what u want, i just want it to work as intended.

KingOfTheFBI:
Thank you king

trende2001:
@Kaikenny can you reopen this please

PR #12, opened 2025-11-26
Only delay FakeLagProcess if no packets were handled

nixxquality:
The fake lag process has a 1ms delay in it to prevent it from running at full tilt when no packets are ready to be handled.

However, if a packet is handled, the 1ms delay is still there. This means that if you have a bunch of packets coming in at the same time, they will be delayed by 1ms each, and probably be cancelled and delayed until the following frames.

This PR only uses the 1ms delay if no work was performed during the current loop iteration.
This resolves https://github.com/Facepunch/sbox-public/issues/1134 and https://github.com/Facepunch/sbox-public/issues/622

https://github.com/user-attachments/assets/c7fe9788-e3ef-444d-8f7c-41278faae350

### DIFF
--- a/engine/Sandbox.Engine/Systems/Networking/Transports/Tcp/TcpChannel.cs
+++ b/engine/Sandbox.Engine/Systems/Networking/Transports/Tcp/TcpChannel.cs
@@ -127,10 +127,13 @@ internal class TcpChannel : Connection
 		{
 			while ( !tokenSource.IsCancellationRequested )
 			{
+				var didSomething = false;
+
 				if ( fakeLagIncoming.TryPeek( out var i ) )
 				{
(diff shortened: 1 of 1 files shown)


engine/Sandbox.Engine/Systems/Networking/Transports/Tcp/TcpChannel.cs
```diff
@@ -127,10 +127,13 @@ private async Task FakeLagProcess()
 		{
 			while ( !tokenSource.IsCancellationRequested )
 			{
+				var didSomething = false;
```
lolleko: Can we have a more descriptive name (e.g. processedPackages / handledPackage).

PR #13, opened 2025-11-26
Allow frozen collections

766974616c79:
Frozen collections can improve my performance since I'm using immutable hashmaps

Resolves https://github.com/Facepunch/sbox-public/issues/730


garrynewman:
No opposed, but we'd need to have a real good look at what we might accidentally be whitelisting along with this.

PR #17, opened 2025-11-26
Fix AssetBrowser view mode persistence and icon update

MrBrax:
Replaces the static backing field for ViewModeType with an instance property to prevent cross-instance issues. Ensures AssetList.ViewMode and the view mode icon are updated correctly when loading settings, addressing problems where the icon would not reflect the current view mode.

### DIFF
--- a/game/addons/tools/Code/Editor/AssetBrowser/AssetBrowser.cs
+++ b/game/addons/tools/Code/Editor/AssetBrowser/AssetBrowser.cs
@@ -109,20 +109,18 @@ public partial class AssetBrowser : Widget, IBrowser, AssetSystem.IEventListener
 		}
 	}
 
-	static AssetListViewMode _viewModeType;
-
 	/// <summary>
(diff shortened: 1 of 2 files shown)


game/addons/tools/Code/Editor/AssetBrowser/AssetBrowser.cs
```diff
@@ -109,20 +109,18 @@ public bool ShowRecursiveFiles
...
 	/// </summary>
 	public AssetListViewMode ViewModeType
 	{
-		get => _viewModeType;
+		get;
 		set
 		{
-			if ( _viewModeType == value ) return;
-			_viewModeType = value;
+			if ( field == value ) return;
```
lolleko: Can you check how often this setter is called? if it is not very often (e.g. every frame) I would just drop that if/guard statement.

MrBrax: It only seems to be set on UI input

PR #18, opened 2025-11-26
Envmap Probe Time Slicing feature

K3rhos:
I had previously requested this feature here: [Envmap Probe Time Slicing](https://github.com/Facepunch/sbox-public/issues/691) but no one really paid attention to it.

So here it is, just like what Unity is doing, instead of rendering all the 6 faces of the cubemap in a single iteration, time slicing spread the rendering of the 6 different textures of a cubemap over 6 frames instead ! (1 frame = 1 face)

Capture d&#39;écran 2025-11-26 001249

This drastically improve realtime envmap probe performance while almost noticing no difference in visual quality !

Here is a quick performance comparison in the default Sweeper scene, with an EnvmapProbe set to Every Frame refresh with and without Time slicing feature enabled:

**Time Slicing Off**
Capture d&#39;écran 2025-11-26 001754

**Time Slicing On**
Capture d&#39;écran 2025-11-26 001734

**Performance Graph**
image

If there is anything you would suggest for me to improve before merging this PR, please let me know.


DevulTj:
Haven't looked at the viability of this, but make sure you run dotnet format because this'll fail without it

K3rhos:
> Haven't looked at the viability of this, but make sure you run dotnet format because this'll fail without it

Make sure to run dotnet format ? What do you mean ?

bub-bl:
Run the `dotnet format` format command in your cmd in your root project directory to follow the same format convention of sbox

K3rhos:
> Run the `dotnet format` format command in your cmd in your root project directory to follow the same format convention of sbox

Okay thx, I didn't know this command existed, that's actually pretty useful, anyway I've done the reformat.

kristiker:
Does update strategy become "Every 6 frames" with this? Maybe it can be named that and the checkbox becomes unnecessary.

K3rhos:
> Does update strategy become "Every 6 frames" with this? Maybe it can be named that and the checkbox becomes unnecessary.

Update strategy still has it's own separated behaviour, update strategy just spread the need to update the probe, its just a signal sent to the probe: "Hey I need to be updated", where time slicing spread the 6 faces refresh itself after it has been marked as "need to be updated", but still with the update strategy taken into account, so update strategy behaviour still makes sense.

As an example if you have Update Strategy set to "Frame Interval", and you set it to 6 + Time Slicing On you will get a spread refresh over 12 frames (6 frames spread of the time slicing + 6 frames for the update strategy)

handsomematt:
How does Unity do this

K3rhos:
> How does Unity do this

Unity has a combo box with like 3 options:
<table class="list">
                  <thead>
                    <tr>
                      <th>Property</th>
                      <th>Description</th>
                    </tr>
                  </thead>
      <tbody><tr><td class="lbl">AllFacesAtOnce</td><td class="desc">Spreads the cubemap update over 9 frames.</td></tr>
      <tr><td class="lbl">IndividualFaces</td><td class="desc">Spreads the cubemap update over 14 frames.</td></tr>
      <tr><td class="lbl">NoTimeSlicing</td><td class="desc">Updates the entire cubemap in one frame. (What S&box is doing right now)</td></tr>
    </tbody></table>
But in my case I've just made a simple checkbox bcs it wasn't useful to spread the faces refresh over more than 6 frames, 6 faces = 6 frames, it was more convenient and you don't really loose any visual quality even when stuff move fast.

Anyway, idk if my approach is the best for that, but at least that's a proposal, if you guys internally have better solution or a better approach I would be glad to see something like that on S&box someday.

AustinDotCodes:
I love the option of it being there, for huge games any ounce of performance that you can get is worth it. and if other engines are doing this, it makes sense to do testing and research if it's worth the update.

Ltbeef:
i'd redo the pr and so you can run unit tests easier

handsomematt:

PR #19, opened 2025-11-26
Normalize decimal separator in FloatControlWidget input parsing

MrBrax:
Replaces commas with periods in input text to handle cultures that use commas as decimal separators, ensuring consistent float parsing from user input.

### DIFF
--- a/engine/Sandbox.Tools/ControlWidget/FloatControlWidget.cs
+++ b/engine/Sandbox.Tools/ControlWidget/FloatControlWidget.cs
@@ -127,7 +127,11 @@ public class FloatControlWidget : StringControlWidget
 	internal static object StringToValueImpl( string text, SerializedProperty property )
 	{
 		Type underlyingType = Nullable.GetUnderlyingType( property.PropertyType ) ?? property.PropertyType;
-		return Convert.ChangeType( text.ToDoubleEval( property.As.Double ), underlyingType );
+		
+		// Normalize comma to period for parsing, some cultures use comma as decimal separator, making the numpad dot input useless
+		string normalizedText = text.Replace( ',', '.' );
+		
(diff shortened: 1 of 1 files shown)


lolleko (changes requested):
Seems like a hack, we should make the float input respect the systems culture.
And decide based on that if it should use a comma or point.

Can probably use something like this:

`Thread.CurrentThread.CurrentCulture`

```cs
var cultureInfo = CultureInfo.GetCultureInfo("de-DE");
double result = Double.Parse("0,5", cultureInfo);
```

MrBrax:
Would mean that the input field should render with a comma too, a slightly more substantial implementation.

lolleko:
Haven't looked into specifics, but wouldn't it be enough to just adjust `ValueToStringImpl` ?

MrBrax:
Doesn't that method only handle input and not rendering? The text in the field has to change too.

MrBrax:
i read your message wrong and confused it with the reverse one. yes, that would be it

garrynewman:
FYI it's very possible that we force `Thread.CurrentThread.CurrentCulture` to en-US or something, because a bunch of stuff was broken in certain conditions in arabic etc

MrBrax:
Only using culture for input parsing wouldn't work, because regardless of my regional settings, if i press "del" on my numpad i get a comma, even though my region settings are set to use a dot. It's dependent on keyboard layout.

handsomematt:
We force the culture, which we probably shouldn't be doing https://github.com/Facepunch/sbox/blob/53ca499e44b5f34cb3fd231b7803e5f5dff9b3ab/engine/Sandbox.Engine/Core/Bootstrap.cs#L43-L48

I think there's a lot of places we need to specify we want to use an invariant culture so that the user culture can be used properly

handsomematt:
Wrong fix, need to fix the culture shit

PR #21, opened 2025-11-26
Expose SoundStream and fix LaughterScore on VoiceComponent

trende2001:
Exposes the SoundStream property on VoiceComponent to allow for getting raw voice data (e.g. playback a player's voice in multiple places at once)

Resolves https://github.com/Facepunch/sbox-public/issues/626

Also LaughterScore has been hooked up to sound.LipSync.LaughterScore as a quick accessor


aylaylay:
Bad design, we don't want to just lazily expose the stream

rodlopt:
WORST DESIGN EVER

PR #22, opened 2025-11-26
Separate Auto Exposure and Tonemapping

rzkid:
A comment in Tonemapping.cs said that the Auto Exposure should be separate, and I agree.
```cs
	//
	// All of this auto exposure stuff should be it's own component
	// It's used by tonemapping, not part of it
	//
```

So, I separated it. That's it. It also works with volumes (why wouldn't it).

Technically satisfies https://github.com/Facepunch/sbox-public/issues/801 even if not directly fixed (this is better)

In theory, having the secondary AutoExposure directly on the camera component is redundant now, but it's not for me to judge whether it is, so tell me if it is, I can add an obsolete in another commit.


kristiker:
I don't think the renderpass where it multiplies `g_flTonemapScalar` is needed. (there should be no shader changes)

rzkid:
Honestly, fair, I was thinking to completely decouple it and making it standalone so it doesnt need anything present to work but it turned out to be more convoluted, you're right

sampavlovic:
Does this break autoexposure in existing components and scenes?

garrynewman:
This doesn't seem worth fucking with, really.

rzkid:
Unfortunately it does break it (by disappearing), as there is literally a separate component now, but after fucking with JsonUpgrader for like an hour, I couldn't make a new component on the same GO out of it, so I can only offer obsoleting the properties :( (at least they tell you there is a new component)

> This doesn't seem worth fucking with, really.

I disagree, the alternative (camera component properties that I complained in an issue barely worked in the first place) does not work with pp volumes, which literally any person who works with auto exposure (more than setting it once on the player prefab) absolutely needs, and this does (plus there was a comment in the tonemap component that this shouldve been done). All engines support exposure volumes or triggers in some form out of the box.

garrynewman:
And you don't want to use the Tonemapping component to do this?

rzkid:
Short answer: 
I have my own component that does things I need it to do, which is separate from the default one. But it cant have the settings on it since the override on camera doesn't like it for some reason (the linked issue from OP)

Long answer:
Having the engine's tonemap component set to linear, alongside my own just so i can have exposure is way weirder, than just having a separate component for this (that still works with volumes!), which is also why I undid the double pass from the initial commit as it was suggested (in retrospect very stupid, but I had my reasons to try it that way so its always there and independent of anything, but thats a kind of unnecessary extra render pass)

And also, you added the camera AE override option
image
which means (i am assuming but its not a long shot) that YOU wrote the comment that says
```cs
	/// <summary>
	/// Enables and configures auto exposure on the camera. This is usually controlled
	/// by the Tonemapping component. But if you're not using that, it can be controlled manually here.
	/// </summary>
```
but then ask "why not" :(
(and this override can't be put on volumes since its a camera AND code only)

sampavlovic:
We really dont want to break existing functionality

rzkid:
I would love there to be a way to technically pull this off, but if even Garry (in discord) said there is probably no way to jsonupgrade into a second component (while keeping self) - idk what can i do here.

Does everything have to suck forever now to keep legacy support? In a dev preview of an unreleased engine?

Sounds oddly familiar.
image
image
(im sorry for pulling the unity card)

aidencurtis:
This is probably too much bloat, but you could add a bool property which is automatically upgraded to be true (is this possible?) called `bool UseDeprecatedAutoExposure`. If this property is true, we show the AutoExposure property group, otherwise it's hidden in editor and unused. This way, only people with preexisting Tonemapping components will see the category, and behaviour will work as expected. Then, you can add the AutoExposure component while maintaining compatability.

Pros:
- Prevents breaking code changes
- Allows for separation of Tonemapping and AutoExposure for new usages, and anyone who wants to clean up their old usage

Cons:
- Adds some bloat in the code for the lifetime of this component (forever)
- Hard to communicate the deprecation succinctly (maybe it'd be nice to have tooltips for property groups in the editor that you could mouse over to see errors / warnings for stuff like this)

MonkeyShark:
> We really dont want to break existing functionality

My project regularly breaks regardless (have to make a private copy of the repo otherwise I need to bug fix weekly). As rzkid said, this a development preview, and since there has been no easy maintained archive system besides the new open source repository, breakage is already part of the process and expected. Better to break things now than have a bloated, ugly engine in the future, or randomly break it when the community is much bigger on release.

I think this is an important change anyway.

handsomematt:
It's my comment, if this makes sense we should do it. It would be easy to make it not break.

handsomematt:

PR #23, opened 2025-11-26
Fix rotated child panels translating in the wrong transform space

Remscar:
Fixes bug https://github.com/Facepunch/sbox-public/issues/645

* In `PanelRenderer.Matrix.cs`, the transform origin is now converted using the parent's inverted global matrix before applying translation, ensuring correct positioning in transformed hierarchies.
* The translation and restoration of the origin now use the transformed origin instead of the raw origin, which aligns panel transformations with the parent's transformed space.

Before:
https://github.com/user-attachments/assets/ef3a9694-4b35-455c-8843-4bc7afbad8b3

After:
https://github.com/user-attachments/assets/e0d356f9-c676-4b54-a89e-329f01c67a63


handsomematt (approved):
Amazing thank you

PR #25, opened 2025-11-26
Fix float slider when using multiple monitors with different DPI scales

gio3k:
Fixes Facepunch/sbox-public#1539

note: Something about Application.CursorPosition seems off here across different monitors - the comments imply it just scales for DPI (compared to UnscaledCursorPosition), is that the case?

PR #26, opened 2025-11-26
Fix scene viewport asset dropping with DPI scaling

gio3k:
Fixes dropping assets on the scene viewport (like dragging a prefab onto the viewport)

I can't find the issue to link to this, I'll hunt it down tomorrow

Without the fix: https://files.smallfi.sh/gio/4830cab/sbox-dev_aNpjSgCxbi.mp4
With the fix: https://files.smallfi.sh/gio/39d1258/sbox-dev_avre1eSW4L.mp4


handsomematt:
Do you think it'd make more sense if we make ScreenPixelToRay multiply by DpiScale

giodotblue:
Hmm... I think that could be a bit confusing, that's part of game code so people might not expect the editor DPI to matter
... it's also a bit confusing without doing that change, so that's a problem

The behaviour [here](https://github.com/Facepunch/sbox-public/pull/25#issue-3668664404) is also adding some confusion, in general the DPI scaling might need some help to be more cohesive

For now, it might be good to add a ScreenPosition (local pos * dpi scale) to DragEvent and then this PR could use that? What do you think?

joshuascript:
Love this. This drives me bananas in the engine.

handsomematt:

PR #27, opened 2025-11-26
Fix Asset Browser clearing on Asset recompile

trende2001:
Removes an unnecessary AssetList.Clear() call in UpdateAssetListAsync.

Resolves https://github.com/Facepunch/sbox-public/issues/829


garrynewman:
This is an awful lot of undocumented code for something that sounds so simple

trende2001:
> This is an awful lot of undocumented code for something that sounds so simple

This fixed my issue but I can simplify it further of course

trende2001:
@garrynewman What about now? I've simplified it to 2 methods, just capture the scroll position and restore it, and documented them.

MonkeyShark:
The problem this fixes is extremely annoying and the code looks pretty good to me.

trende2001:
I would like for this to get re-reviewed.

handsomematt (changes requested):
Why does the asset browser go to the top after a recompile - we should look at that instead of hacking in a position restore

handsomematt (approved):

PR #28, opened 2025-11-26
Property editor changes will call `OnValidate` on components

Remscar:
- When using the undo scope only components specifically listed in `.WithComponentChanges()` will have `Validate` called. Documentation on the undo system states that `GameObjectUndoFlags.Components` captures the _list_ of components, not changes to their contents, while `WithComponentChanges` clearly indicates it will capture changes to the components themselves.
- An edge case arises with prefab instances where `.WithComponentChanges()` is ignored when modifying a prefab instance in the scene. This is why `_capturedComponentsToValidate` is used to keep track of all components passed from the builder via `WithComponentChanges`
- Tested: Undo and Redo properly call `OnValidate` when a value changes
- In testing I discovered that pasting a component "As New" will not call `OnValidate()` so I fixed that as well (Change to `Component.DeserializeImmediately`)

There is a potentially sub-optimal behavior that needs to be fixed:

**"Revert Property Instance Change" calls `OnValidate` twice**
- Once from `Component.DeserializeImmediately` which `EditorUtility.Prefabs.RevertPropertyChange` ultimately calls
- Once from the UndoSnapshot (via the code that was introduced in this PR)

If this is a blocker then this behavior could be fixed be adding a `validate` flag to `Component.DeserializeImmediately` to conditionally call OnValidate. `Component.DeserializeImmediately` seems to only ever be called by Prefab related code.


lolleko:
Unfortunately, this is not sufficient. The Inspector is not the only way a component can be edited. This should be hooked into the undo system, which is a bit more complicated.

Remscar:
> Unfortunately, this is not sufficient. The Inspector is not the only way a component can be edited. This should be hooked into the undo system, which is a bit more complicated.

Good point, thanks for the feedback.

The `SceneUndoSnapshot.ComponentSnapshot` undo system will now call `OnValidate` on components, after `PostDeserialize` to respect the same order observed in `CommonCallback`

engine/Sandbox.Tools/Scene/Session/SceneEditorSession.Undo.cs
```diff
@@ -210,6 +210,15 @@ public void PostRestore( Scene scene )
 					compRef.Resolve( scene )?.PostDeserialize();
 				}
 			}
+
+			foreach ( var compRef in ComponentRefs )
+			{
+				var comp = compRef.Resolve( scene );
+				if ( comp.IsValid() )
+				{
+					comp.Validate();
```
Remscar, line 219: Unsure if the above foreach loop doing `compRef.Resolve( scene )?.PostDeserialize();` when it already has a non-null `comp` was intentional, but I went with the simpler form here.

lolleko (changes requested):
Thanks for giving this a try we still need some more changes though.
Sorry maybe i wasn't clear enough, `OnValidate` should only be invoked by the undo system, not via widget.
It should be invoked:

1. When a component snapshot is created (this happens after a change via gizmo, inspector change, context menu change).
2. When undo is invoked and a component has it's state restored
(might not be needed since Component.Deserialize already calls on Validate)
3. When redo is invoked and a component has it's state restored
(might not be needed since Component.Deserialize already calls on Validate)

There is also the case where component changes are captured in the undo system by storing the whole GameObject `WithGameObjectChanges( go, GameObjectUndFlags.Components)` or `WithGameObjectChanges( go, GameObjectUndFlags.All)`.
I am not sure what do do in this case. Maybe we can just run `OnValidate` on all components of the Go in that case.
(might not be needed since Component.Deserialize already calls on Validate)

You should get rid of the invocation in the ComponentListWidget and only do it in the correct places in the undo system.

And finally: instead of using `EditorEvents` we usually put these helpers into `EditorUtility`.

To test this, add a log to on validate and then verify it is triggered when modifying a component via gizmos (e.g. box collider), when its changed via context menu and via inspector.

engine/Sandbox.Tools/Scene/Session/SceneEditorSession.Undo.cs
lolleko: Probably just an oversight , feel free to simplify the above loop.
I also think this could be just one loop.
Unless you want to make sure changes to all components are restored before calling OnValidate.
Honestly no strong opinion on this,

Remscar: When components are loaded in on game object deserialization all of them are first PostDeserialized before Validate, so I think it should be the same in this case.

Remscar:
Okay here is the next revision;

- When using the undo scope only components specifically listed in `.WithComponentChanges()` will have `Validate` called.
- An edge case arises with prefab instances where `.WithComponentChanges()` is ignored when modifying a prefab instance in the scene. This is why `_capturedComponentsToValidate` is used to keep track of all components passed from the builder via `WithComponentChanges`
- Tested: Undo and Redo properly call `OnValidate` when a value changes
- In testing I discovered that pasting a component "As New" will not call `OnValidate()` so I fixed that as well (Change to `Component.DeserializeImmediately`)

There is a potentially sub-optimal behavior that needs to be fixed:

**"Revert Property Instance Change" calls `OnValidate` twice**
- Once from `Component.DeserializeImmediately` which `EditorUtility.Prefabs.RevertPropertyChange` ultimately calls
- Once from the UndoSnapshot (via the code that was introduced in this PR)

If this is a blocker then this behavior could be fixed be adding a `validate` flag to `Component.DeserializeImmediately` to conditionally call OnValidate. `Component.DeserializeImmediately` seems to only ever be called by Prefab related code.

@lolleko You mention that "Component.Deserialize already calls on Validate" but I don't see that anywhere in the codebase. I only ever see OnValidate being called on Components if the game object is deserialized. Making components call Validate could be a more elegant solution than what i've implemented for the redo/undo, if you would prefer that let me know.

lolleko:
That's what i meant with it will be a bit more complicated. Thanks for hanging in.

I am a bit worried to touch `DeserializeImmediately` it's a public method changing it, may change behavior games rely on.
Maybe instead of adding a OnValidate call there you call it manually after "paste as new". This would also fix the issue of it being called twice in other instances.

Not a big fan of `_capturedComponentsToValidate`, but also don't see a better way right now, could you add a comment explaining why it exists?

I think the whole way how we handle component deserialization PostSerialize, OnValidate needs a make over.
But that's out of scope for this PR for now I would be happy with this for now and validate being called on editor changes again.

===

That being said the longer I think about OnValidate, I don't think it should exist.

We have `[Validate( nameof( IsValid ), "Warning Message", LogLevel.Warn )]` for properties which can handle editor validation.

I don't know how much `OnValidate` is used in game code, but it seems adding a callback called `PostSerialize` might make more sense.

Remscar:
`_capturedComponentsToValidate` is added because when an UndoScope is created `.WithGameObjectChanges(x, GameObjectUndoFlags.Components)` then any components listed in `.WithComponentChanges(x)` are not actually included in `_captureComponentChanges `.

We need to know what components are included in the snapshot to be able to call `OnValidate` on them.

Another case is with prefab instances; where `_captureComponentChanges` is never set for components in the prefab.

--

Just using a `[Validate]` decorator, removing `OnValidate` entirely and replacing it with a `OnDeserialize` could be a good solution. I personally prefer having a global sort of "Property validation" callback compared to having to decorate every property, and i'd be concerned about breaking anything using `OnValidate`.

We can abandon this PR though if you'd rather tackle this at a later date, I know this is pretty inconsequential. A simple change to the docs saying "OnValidate" is only called after deserialization would be an immediate bug fix too :)

PR #30, opened 2025-11-26
Add AddRange support to NetList

Matt9440:
**Old behaviour**
```
[Sync] public NetList<string> MyList { get; set; }

public void OldBehaviour()
{
	var valuesToAdd = new string[] { "thing 1", "thing 2" };

	MyList.ToList().AddRange( valuesToAdd );
}
```

**New behaviour**
```
[Sync] public NetList<string> MyList { get; set; }

public void NewBehaviour()
{
	var valuesToAdd = new string[] { "thing 1", "thing 2" };

	MyList.AddRange( valuesToAdd );
}
```

### DIFF
--- a/engine/Sandbox.Engine/Scene/Networking/Containers/NetList.cs
+++ b/engine/Sandbox.Engine/Scene/Networking/Containers/NetList.cs
@@ -161,7 +161,7 @@ public sealed class NetList<T> : INetworkSerializer, INetworkReliable, INetworkP
 	/// <summary>
 	/// <inheritdoc cref="List{T}.AddRange"/>
 	/// </summary>
-	public void AddRange( T[] collection )
+	public void AddRange( IEnumerable<T> collection )
 	{
 		if ( !CanWriteChanges() )
 			return;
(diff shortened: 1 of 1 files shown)


engine/Sandbox.Engine/Scene/Networking/Containers/NetList.cs
```diff
@@ -158,6 +158,20 @@ public void Add( T value )
 		list.Add( value );
 	}
 
+	/// <summary>
+	/// <inheritdoc cref="List{T}.AddRange"/>
+	/// </summary>
+	public void AddRange( T[] collection )
```
garrynewman: Is it normal for this to be T[] instead of IEnumerable<T> or something

Matt9440: I initially had it as IEnumerable<T> but followed suit with the CopyTo method which uses T[]. I'm happy to change it?

Matt9440: Actually, it makes more sense for this to be an IEnumerable.

PR #35, opened 2025-11-27
Fix CheckValidationAttributes call within ReviewWizardPage not outputting all error messages

TrvsF:
i was getting a silent error when trying to publish a project with a Title above 32 characters
this change makes any errors that occur when Project.Config contains invalid properties appear to the user

image


Madmegsox1:
looks like a very useful pr, i think Ive gotten stuck here once or twice...

PR #38, opened 2025-11-27
Update MapInstance to support scene maps

zonical:
This PR is a band-aid solution that allows the loading of scene maps via the MapInstance component. The changes are very minimal, and my map Street loads perfectly fine.

> 💬 MapInstance sucks, we should get rid of it!

Yes, yes, yes, I've heard that MapInstance is the devil and that Facepunch wants to get rid of it at some point. I'm all for that, but there's no established timeline for when scene maps will be loadable and usable in-game/in-editor.

> 💬 Why create your own component to load the scenes?

I have tried this, but I've run into a bug where a `SceneWorld` created from a `SceneMap` class doesn't get removed properly. This bug can be prevented, but the change must be made at a lower level that normal components can't access. MapInstance appears to clean things up properly with the appropriate level of access, so that's why I've put these changes in here.

image

_Example of the bug in action. This is out of the scope for this PR and should be addressed separately._

### DIFF
--- a/engine/Sandbox.Engine/Scene/Components/Map/MapInstance.cs
+++ b/engine/Sandbox.Engine/Scene/Components/Map/MapInstance.cs
@@ -44,8 +44,7 @@ public partial class MapInstance : Component, Component.ExecuteInEditor
 	SceneMap loadedMap;
 	GameObject _mapPhysics;
 	string loadedMapName;
-
-
+	string sceneMapScenePath;
 
 	public MapInstance() : base()
 	{
@@ -119,6 +118,7 @@ public partial class MapInstance : Component, Component.ExecuteInEditor
 	public void UnloadMap()
 	{
 		loadedMapName = null;
+		sceneMapScenePath = "";
 
 		bool hadMap = loadedMap is not null;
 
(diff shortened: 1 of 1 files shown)


garrynewman:
I thought this already worked. Long term we do need a better solution than MapInstance, but I agree that we should be able to load scene maps through this in the meantime.

zonical:
A check was deliberately added to prevent scenes from being loaded, hence the pull request. I was pleasantly surprised to find how well the MapLoader class actually handled the loading of scene maps during my initial experimentation.

Although this doesn’t fit the purpose of the pull request, it might be worth considering a plan to introduce scene maps in the editor sooner rather than later. It would be really handy if developers could load scene maps and then insert their own GameObjects on top. My game, Physbox, uses a system similar to this where breakable props and other map features such as clip brushes and barricades are manually placed on top of the Map Instance. Just some food for thought.

garrynewman:
We have a couple of things to facilitate that. We have a System Scene - which is auto added to every loaded scene. And we have GameObjectFlags.DontDestroyOnLoad - which will make GameObjects stay when the scene is changed.

zonical:
System Scenes and GameObjectFlags don't provide the functionality that I'm looking for. The approach I would like to see would be similar to [how Map Extensions are handled in Half-Life: Alyx.](https://developer.valvesoftware.com/wiki/Half-Life:_Alyx_Workshop_Tools/Level_Design/Map_Extensions) More specifically, treating scene maps as just another scene that can be loaded in the editor (the functionality is technically there since scene maps are just `SceneFile` resources, IIRC), but in a "read-only" state. The developer can add their own GameObjects on top to extend the original scene. I don't know if that sounds too complicated or if it's at all practical to implement.

This is already _kind of_ possible by creating a new scene, adding a Map Instance component with a Hammer/Scene Map, and then adding your own GameObjects. This is the approach I use in Physbox (see image below).

image

This discussion is probably getting a bit out of scope for this PR. The ability to load scene maps with the Map Instance component is a good temporary measure that can be revisited later.

PR #45, opened 2025-11-27
Reimplement impact damage system for Rigidbody

Asphaltian:
Brings back impact damage in Rigidbody from the old entity system.

Props and other physics objects can now take and deal damage when colliding at high speeds. Damage scales with collision velocity and can be configured per-object.

**New Properties:**
- `EnableImpactDamage` - Toggle impact damage (default: true)
- `MinImpactDamageSpeed` - Speed threshold to trigger damage (default: 500)
- `ImpactDamage` - Custom damage value, or leave at 0 to auto-calculate from mass

**Demo:**

https://github.com/user-attachments/assets/ea3c095a-3ec0-49d5-996b-37d74c978bda


engine/Sandbox.Engine/Scene/Components/Game/Prop.cs
```diff
@@ -546,6 +565,69 @@ public List<Gib> CreateGibs()
...
+
+		// This prop takes damage from high speed impacts
+		if ( Health > 0 )
+		{
+			var selfDamage = damageMultiplier * impactDmg;
+			var damageInfo = new DamageInfo( selfDamage, collision.Other.GameObject, collision.Other.GameObject )
+			{
+				Position = collision.Contact.Point
+			};
+			damageInfo.Tags.Add( "physics_impact" );
```
garrynewman: Lets change this tag to physicsimpact, or ideally just impact

engine/Sandbox.Engine/Scene/Components/Game/Prop.cs
```diff
@@ -546,6 +565,69 @@ public List<Gib> CreateGibs()
...
+
+			OnDamage( damageInfo );
+		}
+
+		// The other object takes more damage
+		var otherDamageable = collision.Other.GameObject?.GetComponentInParent<IDamageable>();
```
garrynewman: If the other object is a prop, will OnCollisionStart get called for that one too, resulting in double damage?

garrynewman (commented):
I wonder if this is a prop thing or a Rigidbody thing? Does it make sense that only props can create impact damage?

Asphaltian:
> I wonder if this is a prop thing or a Rigidbody thing? Does it make sense that only props can create impact damage?

Looking more closely at the original Prop entity and I'm having the same thoughts, figure I'll move this to Rigidbody and use a simpler tag instead

Also yes @garrynewman, the double damage is intended since that's how it worked before (can be changed ofc)

PR #48, opened 2025-11-27
Display s&box version in editor

nixxquality:
After switching back and forth between working on sbox-public and the Steam version, I find myself uncertain which version of the editor I'm currently using.
In game this is easy to see, there's a blue tooltip which shows the version at all times, but I couldn't easily find it in the editor.

This patch adds it to two places in the editor (in these screenshots I've copied over the `.version` file from the Steam release, usually it's `0000000` in sbox-public).

Editor startup log:

About page:

If you dislike either of these I can change/remove it.


lolleko:
sure why not

PR #54, opened 2025-11-27
Show cookie name when expiring a single one

MrBrax:
The log message you get it very vague, 1 cookie expired. This makes it show what it actually is. Rarely is there more than one cookie expiring at one time.


garrynewman:
I don't think we even need to print anything, this is just left in debug shit

PR #56, opened 2025-11-27
Min Length revision for library / package

766974616c79:
You can create a library with a single character, but the library will not be uploadable and will create errors (due to the API response). 
You can also create a library or package with a title size of 2, but the API does not allow any changes if the title size is less than 3.
My PR requires at least two characters when creating the library. And I increment the MinLength of the title.

Resolves https://github.com/Facepunch/sbox-public/issues/1253


internetfishy:
I believe the website API requires 3 characters? I have an issue open where I tried to update a 2 character one.

https://github.com/Facepunch/sbox-public/issues/1253

766974616c79:
I increased the MinLength of the title to 3.

PR #57, opened 2025-11-27
Extend ISceneEvent RunEvent to support UI panels in Scene

MrBrax:
The RunEvent<T> method in Scene now also runs the event on all PanelComponent panels and their descendants that implement the event interface, in addition to components.

Implements https://github.com/Facepunch/sbox-public/issues/3359

I am VERY unsure about the performance implications of this, looping through all descendants could be expensive.

### DIFF
--- a/engine/Sandbox.Engine/Scene/Scene/Scene.Event.cs
+++ b/engine/Sandbox.Engine/Scene/Scene/Scene.Event.cs
@@ -1,9 +1,11 @@
-﻿namespace Sandbox;
+﻿using Sandbox.UI;
+
+namespace Sandbox;
 
 public partial class Scene : GameObject
 {
 	/// <summary>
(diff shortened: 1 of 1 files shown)


MrBrax:
I had a thought of doing an early return if the passed T event wasn't an ISceneEvent, as rendering and editor events called each frame has no need to run on panels

MrBrax:
Could also just create a new IPanelEvent interface for panel events only, so they're not clobbered by other events

garrynewman:
This is going to be so slow

MrBrax:
Yeah, fair. What about implementing a new event interface just for panels so they aren't called all the time?

handsomematt (changes requested):
Register the interfaces somewhere for panels the same way it's done for components, then fetch them and execute, no recursion.

MrBrax:
no clue how that would work, i'll leave this one

PR #63, opened 2025-11-28
Removed IsUnitTest check in Material Load functions

TheDimin:
I have code that i want to test which contains map with material as the key.
unit tests fail as it would always return null.

After some testing it seems that material loading does work when running unit tests.
So removing these checks should be fine?

note that it will just return null if you don't call `Sandbox.Application.InitUnitTest<TestClass>( true,true );`
But that seems logical as Materials are tied to rendering


garrynewman:
Worth a try

PR #65, opened 2025-11-29
Allow for getting and setting of static properties in TypeLibrary

Nolankicks:
Resolves https://github.com/Facepunch/sbox-public/issues/3652


garrynewman:
Any anticipated side effects? Why do you think it was testing for static before?

Nolankicks:
I mean there was a workaround to still get static properties and set them. They seemed to behave fine in my testing. Please lmk if there are any issues with them being static and I’ll make sure to fix it.

Nolankicks:
I see the issue with refactoring the way I did it. I just pushed an alternative solution. Please lmk @garrynewman

garrynewman:
I don't totally hate this

PR #68, opened 2025-11-29
Add ability to configure custom Code Editors

scomitch:
## Summary
Adds support for configuring custom code editors, addresses https://github.com/Facepunch/sbox-public/issues/504

## Why
Currently, S&Box only supports a fixed set of editors (Visual Studio, VSCode and Rider). Users who prefer to use any other editor outside of these three are essentially blocked from utilizing any of the built in functions to open files inside of the S&Box editor.

## Features
- Add any editor executable you want
- Custom argument templates for any editor {path}, {line}, {column}, {solution} args
- Settings menu to change arguments, delete editors etc.
- For UX, any editors not found will state that instead of just disabling clicks.
- Persistent storage, utilizes EditorCookie to store configurations.

## Screenshots

sbox-dev_2025-11-29_19-50-05
sbox-dev_2025-11-29_19-50-38
sbox-dev_2025-11-29_19-50-52
sbox-dev_2025-11-29_19-51-01
sbox-dev_2025-11-29_19-51-17

https://github.com/user-attachments/assets/a181bd6f-7cd9-47cd-bffe-cd4deee151e7


BumpyJumpy120:
tbh this is actually something good, hope they add this

Remscar:
This would be so nice

PR #69, opened 2025-11-29
Fix stack trace parsing for error messages containing "in"

scomitch:
## Summary
Clicking on certain compile errors in the console fails with "Couldn't resolve relative path" when trying to open them in the editor.

## Cause
Non-greedy matching issue, so any error message with the word **"in"** in it, breaks the flow and parses the wrong path when going to open the editor.

## Fix
Update the regex to actually require the path to be absolute.

## Screenshots

(Before fix)
image

(After fix)
image


garrynewman:
This is your code @Metapyziks - does this look right?

Metapyziks:
Makes sense, I'll see if it's feasible to add tests for this

Metapyziks (changes requested):
Let's add some tests in Sandbox.Test/Editor/StackTrace.cs like this:

```csharp
using System.Reflection;
using System.Text.RegularExpressions;
using Editor;

namespace TestEditor;

[TestClass]
public class StackTraceTests
{
	[TestMethod]
	[DataRow( @"at Example.Foo() in C:\sbox\Source.cs:line 123", true )]
	[DataRow( @"cat Example.Foo() in C:\sbox\Source.cs:line 123", false )]
	[DataRow( @"at Error in something in C:\sbox\Source.cs:line 123", false )]
	public void DefaultStackLine( string line, bool expectMatch )
	{
		var isMatch = GetDefaultStackLineHandlerRegex().Match( line );

		Assert.AreEqual( expectMatch, isMatch.Success );
	}

	private Regex GetDefaultStackLineHandlerRegex()
	{
		var method = typeof( StackTraceProperty ).GetMethod( nameof( StackTraceProperty.DefaultStackLineHandler ), BindingFlags.Static | BindingFlags.Public )!;

		return new Regex( method.GetCustomAttribute<StackLineHandlerAttribute>()!.Regex );
	}
}
```

scomitch:
Will get the build passed after this weekend, traveling and using a hellfire that is the macbook so wont be fixed until i get home

Metapyziks:
No worries, thanks for your help!

PR #77, opened 2025-12-01
Fix Network Refresh overwriting properties marked with FromHost SyncFlag

andy013:
Issue: https://github.com/Facepunch/sbox-public/issues/564

### Changes

- Skip `[Sync]` properties in `PostDeserialize` during a network refresh.
- Check if connection has control when reading data table after a refresh

### Notes

This fix prevents clients who call a network refresh from overwriting `Sync` properties that have the `FromHost` SyncFlag.

Currently, `[Sync]` properties are deserialised from the client's JSON payload, overwriting authoritative host values. `ReadDataTable` also lacked a check, causing `NetList` and `NetDictionary` types to be reset.

Checks are performed on the receiver side.  I figured you should always have the receiver check if the sender was allowed to change the value to prevent hacked clients sending data and it just being blindly accepted. You could probably optimise it by also not sending `FromHost` values on a refresh called by clients.

I've not tested this extensively but it seems to be working in my basic tests. You should review it and make sure I'm not doing anything stupid. 

### Example

Test project: [fromhostbug.zip](https://github.com/user-attachments/files/23844663/fromhostbug.zip)

Join with a client and observe server console logs. Client calls `Network.Refresh` every frame which overwrites the values set by the host.

`OnStart()` sets the `Sync` `FromHost` properties like this:
```csharp
protected override void OnStart()
{
	if ( Networking.IsHost )
	{
		Ammo = 6;
		NetworkList.Add( 1 );
		NetworkList.Add( 2 );
	}
}
```

### Before fix:
```
[Generic] Wilson [76561197988413004] is connecting	
[Generic] [HOST] Ammo: 0 -> 1	    <--  Ammo set to 1 from the prefab
[Generic] Wilson (2) [76561197988413004] is connected	
[Generic] [HOST] Ammo: 1 -> 6	    <--  OnStart called for host
[Generic] List Changed: 	
[Generic] 1	
[Generic] 2	
[Generic] [HOST] Ammo: 6 -> 1	    <--  Client refresh sets ammo back to prefab value!
[Generic] List Changed: 	
[Generic] List is now empty.        <--  NetList also reset!
```

### After fix:
```
[Generic] Wilson [76561197988413004] is connecting	
[Generic] [HOST] Ammo: 0 -> 1	
[Generic] Wilson (2) [76561197988413004] is connected	
[Generic] [HOST] Ammo: 1 -> 6	
[Generic] List Changed: 	
[Generic] 1	
[Generic] 2	     <-- Both Ammo and NetList stay set to host values even after client calls Network.Refresh
```


garrynewman:
Does this make sense to you @kurozael?

kurozael:
Makes sense - but I feel like the host should always be able to overwrite Sync Var values in a Network Refresh. So when reading the data table, I'd accept entries if the source is the host *or* we the source is the owner of the Sync Var.

andy013:
Thanks for the feedback. I've updated the code so only client refreshes now skip `[Sync]` properties in JSON. Host refreshes deserialize everything, allowing the host to update all properties while preventing clients from overwriting `FromHost` values.

I also fixed a small bug I found while testing. In `NetworkTable.Register` the lists for `_reliableEntries`, `_snapshotEntries` and `_queryEntries` were never being cleared. This meant that every time a network refresh happened, duplicate entries were added for the same slot and they would just keep accumulating. `Register` now calls `Unregister` first to prevent this.

engine/Sandbox.Engine/Scene/Networking/NetworkTable.cs
```diff
@@ -130,6 +130,8 @@ public void Unregister( int slot )
 	/// </summary>
 	public void Register( int slot, Entry entry )
 	{
+		Unregister( slot );
```
kurozael: Why?

engine/Sandbox.Engine/Scene/Networking/NetworkObject.cs
```diff
@@ -677,7 +677,7 @@ internal void UpdateFromRefresh( Connection source, byte[] tableData, byte[] sna
 		var system = SceneNetworkSystem.Instance;
 		if ( system is null ) return;
 
-		ReadDataTable( tableData, ( _, entry ) => entry.HasControl( source ) );
+		ReadDataTable( tableData, ( _, entry ) => source.IsHost || entry.HasControl( source ) );
```
kurozael: This is the thing that makes sense.

engine/Sandbox.Engine/Scene/Components/Component.Serialize.cs
```diff
@@ -138,6 +138,14 @@ internal void PostDeserialize()
 					continue;
 				}
 
+				// Client refresh
+				if ( options.IsNetworkRefresh && !options.IsRefreshFromHost )
+				{
+					// Skip [Sync] vars - sent via network table (prevents FromHost overwrite)
+					if ( field is PropertyDescription prop && prop.HasAttribute<SyncAttribute>() )
```
andy013, line 145: Should I use `ReflectionQueryCache.SyncProperties` instead of `prop.HasAttribute` here?

engine/Sandbox.Engine/Scene/GameObject/GameObject.Serialize.cs
```diff
@@ -107,6 +107,12 @@ public struct DeserializeOptions
 		/// </summary>
 		internal bool IsNetworkRefresh { get; set; }
 
+		/// <summary>
+		/// If true, the network refresh originated from the Host,
+		/// so we should allow it to overwrite [Sync] properties.
+		/// </summary>
+		internal bool IsRefreshFromHost { get; set; }
```
lolleko, line 114: We hould really avoid adding new flags to options whenever possible. Is there no other way?

PR #81, opened 2025-12-02
Fix: Properly destroy Scene in MapWorld to prevent memory leak

PatrickJnr:
Fixes a `FIXME` in [MapWorld.cs](cci:7://file:///t:/sbox-public/engine/Sandbox.Tools/MapEditor/MapDoc/Nodes/MapWorld.cs:0:0-0:0) where the [Scene](cci:1://file:///t:/sbox-public/engine/Sandbox.Engine/Scene/Scene/Scene.cs:80:1-80:35) wasn't being properly destroyed.

Previously, the reference was set to `null`, which meant the Scene (and its native resources) hung around until the GC decided to run. This change ensures `Scene.Destroy()` is called immediately when the MapWorld is destroyed.


handsomematt (approved):
Makes sense, thank you

PatrickJnr:
Thanks @handsomematt

PR #84, opened 2025-12-02
SceneModel changes

yuberee:
Some minor grievances I've been running into for SceneModel

- ~~Make `ModelChanged` public to subscribe to~~
- Make `SceneModel.GetBoneVelocity()` and `SceneModel.HasBoneOverrides` public
- Add `SkinnedModelRenderer.GetBoneVelocity()`
- Document methods, fix some oversights

### DIFF
--- a/engine/Sandbox.Engine/Scene/Components/Render/ModelRenderer.cs
+++ b/engine/Sandbox.Engine/Scene/Components/Render/ModelRenderer.cs
@@ -196,7 +196,10 @@ public partial class ModelRenderer : Renderer, ExecuteInEditor, ITintable, IMate
 		OnModelChanged();
 	}
 
-	internal Action ModelChanged;
+	/// <summary>
+	/// Invoked when <see cref="Model"/> has been changed.
+	/// </summary>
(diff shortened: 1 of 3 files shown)


engine/Sandbox.Engine/Scene/Components/Render/ModelRenderer.cs
```diff
@@ -196,7 +196,10 @@ internal void OnModelReloaded()
 		OnModelChanged();
 	}
 
-	internal Action ModelChanged;
+	/// <summary>
+	/// Invoked when <see cref="Model"/> has been changed.
+	/// </summary>
+	public Action ModelChanged;
```
garrynewman: Nope

yuberee: I was looking for ways to check when the model was changed so I can rebuild my component and found that unused action.
ModelPhysics does it by having a separate SkinnedModelRenderer property and Model property with a set; that calls the rebuild, is that the correct way of doing it?

PR #85, opened 2025-12-03
Fix tools materials in Hammer Level Editor's fullbright mode

Aleph0x:
Resolves https://github.com/Facepunch/sbox-public/issues/625 by adding specular to all tool materials.

This change enables specular on all tool materials using `generic.shader`, ensuring the Fullbright path always has a non-zero specular term to work with. The result is that tool surfaces become readable again in Hammer's fullbright without altering how they behave in-game or after a map compile. Their material definitions are also updated to the current `generic.shader` layout, replacing legacy texture bindings with the modern PBR slots.

Hammer’s Fullbright mode uses `ToolsVis.hlsl`.

`vColor.rgb = vAlbedo * flAmbientCubeFactor * flAmbientCubeAmount + flHeadlightSpec.xxx;`

Many tool materials currently have their `vAlbedo` (LightSim_Albedo) effectively set to `[0 0 0]`. Since Albedo is zero, multiplying it by the ambient cube factor still gives zero:

`vColor.rgb = 0 + flHeadlightSpec.xxx;`

This means the ambient contribution is always black, and the only thing making these materials visible in Fullbright is the tiny baked-in `flHeadlightSpec` term. When specular is disabled on the material, even that small contribution disappears, causing the surface to render almost completely black in the viewport.

This PR also updates their texture bindings. Many tool materials were still using legacy auto-generated bindings like `g_tColor` and `g_tNormal`. Recompiling them under the modern shader layout switches them to `TextureColor`, `TextureNormal`, and the engine’s default PBR slots.

Image

Image


Arcevalis:
Finally. This was getting annoying. +1

handsomematt:
What do you think @sampavlovic, does this make sense or should we be fixing it in ToolsVis.hlsl

PR #87, opened 2025-12-03
Be carefull null asset

766974616c79:
When we save a scene/prefab in a folder that is not our project folder, we get an NRE. My PR uses the result of `CreateResource` to ensure that we can continue.


766974616c79:
https://github.com/Facepunch/sbox-public/commit/f19a719ece8955a4a10827c2f59ca97a642c6cee

PR #89, opened 2025-12-03
Fixes for inline editor in collections

boxrocket6803:
fixes https://github.com/Facepunch/sbox-public/issues/1666 and https://github.com/Facepunch/sbox-public/issues/2123

Dictionary keys no longer get stretched when using larger InlineEditor values, and it's no longer attempting to add a property name label for inline editor properties that are part of a collection (and thus don't have a name label)

image
image
image
(new pr cause it wasn't on a branch and something was messed up with the diff)


Remscar:
This would be so nice, I end up having to create a component which is my dictionary "value" and using the editor on the component itself because the current dictionary editor is so tiny.

boxrocket6803:
@Remscar This doesn't make anything bigger, you're looking for the WideMode attribute

garrynewman:
Everything here looks just as terrible

boxrocket6803:
perhaps a more notable improvement in a less busy scenario
before
image
after
image

PR #90, opened 2025-12-03
Only create Prop sometimes, only add physics sometimes

boxrocket6803:
fixes https://github.com/Facepunch/sbox-public/issues/1579

ModelDropObject always spawned Prop if the model had physics (e.g. every model) and made a fairly flimsy attempt to guess if the model should be static. This resulted in entire scenes falling into the void if the user didn't manually disable physics for each prop dragged into the scene or manually set the archetype of every imported model. This code adds a bit more leeway as to which archetypes are allowed and will default to static if archetype isn't set. It also avoids creating the Prop component entirely if the model has no prop data or break pieces and doesn't have its archetype set to a prop related option.

https://github.com/user-attachments/assets/f4b8fa84-c54a-4a7a-8d1d-b9227b32400f


garrynewman:
I can see the logic here but I think by trying to guess what is desired we're always going to give someone what they don't want. Maybe we should be doing something like if you hold Ctrl and drag it's a renderer, if not, it's a prop?

boxrocket6803:
Main decision is whether or not the prop should be static, from there we can easily determine whether a Prop component is needed based on if the model isn't static and/or has prop data. Still seemed like since model archetype exists it should be used, for asset libraries where it's consistently set (like all the facepunch assets) the code will reliably determine if a vmdl is intended as a physics prop. In cases where archetype is never set it will assume props to be static. Shift (ctrl enables snapping) key inverts that. I also made the bbox blue to indicate that a model will be created as a prop with physics, maybe needs a better indicator for that?

https://github.com/user-attachments/assets/3e49e428-309a-43d4-ba86-0d917b9fd392

I also tried ignoring archetype completely and just spawning with physics unless the user holds shift but it feels kinda uncomfortable to build levels while holding shift all the time (levels are made out of props with collision, you'd have to hold shift while dragging in every model, if you ever miss one part of your scene falls into the void which was the whole issue to begin with). Feels a lot nicer to me to let a bit of effort setting archetype on custom vmdls make everything easier forever onwards than to require an extra key all the time just in case there's a vmdl that's set up wrong

PR #91, opened 2025-12-03
Nicer order when duplicating objects

boxrocket6803:
fixes https://github.com/Facepunch/sbox-public/issues/973

When duplicating objects, rather than arrange each new copy individually under its source create 'groups' from unbroken chains of objects and place new objects in order under their respective group. This helps keep the order of objects consistent when duplicating and also helps keep the rough correlation between time the object was created and position in the list intact where the existing system tends to create fairly chaotic ordering.

Applies to shift+drag in viewport, the ctrl+d shortcut, and the duplicate option in right click menus

image

before, after
image image

### DIFF
--- a/game/addons/tools/Code/Scene/SceneEditorMenus.cs
+++ b/game/addons/tools/Code/Scene/SceneEditorMenus.cs
@@ -20,18 +20,28 @@ public static class SceneEditorMenus
 	{
 		using var scope = SceneEditorSession.Scope();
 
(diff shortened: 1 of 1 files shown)


lolleko:
What happens when you have "gaps" in your selection and duplicate?
(e.g. you selected spline_1, spline4, and then duplicate)

boxrocket6803:
the new objects are created under unbroken chains of selected objects
image

DrakeFruit:
this is awesome thank you

PR #96, opened 2025-12-06
Fix: Clarify Attributes property caching in SceneObject

PatrickJnr:
### Problem
The [Attributes](cci:2://file:///t:/sbox-public/engine/Sandbox.Engine/Systems/Render/RenderAttributes.cs:18:0-393:1) property had a FIXME about allocating `CSceneObject::m_pExtraData` even when only reading. Although the `??=` operator cached the value, the behaviour wasn't explicit.

### Solution
Changed to an explicit `if (_attributes == null)` check for clarity and removed the misleading FIXME comment.

### Testing
The test file was removed before committing.  
Verified that multiple accesses to [Attributes](cci:2://file:///t:/sbox-public/engine/Sandbox.Engine/Systems/Render/RenderAttributes.cs:18:0-393:1) return the same cached instance.

image

### DIFF
--- a/engine/Sandbox.Engine/Systems/SceneSystem/SceneObject.cs
+++ b/engine/Sandbox.Engine/Systems/SceneSystem/SceneObject.cs
@@ -29,8 +29,11 @@ public partial class SceneObject : IHandle
 		get
 		{
 			if ( native.IsNull ) return null;
-			// FIXME: What really sucks with this is it allocates CSceneObject::m_pExtraData even if we're only reading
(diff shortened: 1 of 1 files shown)


chrisspieler:
Regarding the removed comment: 
- How did you determine that the comment was inaccurate?

Regarding the change to null checking: 
- Is the justification for this change that someone reading the source code of the engine might not know that a null-coalescing assignment evaluates to the new value of its left-hand side?

PR #99, opened 2025-12-07
Add formatting to indicate state of sound preview toggle buttons

boxrocket6803:
fixes https://github.com/Facepunch/sbox-public/issues/918

Loop and Auto-Play buttons in sound preview previously had no indication as to their state, this adds a stylish blue background to those buttons when they are enabled
image


joshuascript:
Great tooltip icons.

handsomematt (changes requested):
Really bad way to do it, making a Bitmap and setting that as the new icon in Toggled...

It's nice but it'd need to be a common button style on the Toolbar.AddOption button, and override Paint instead

boxrocket6803:
Seemed like Toolbar and Option are all in native qt and a pain to mess with. Noticed I could just replace them all with IconButton instead though, feels cleaner
image

PR #101, opened 2025-12-07
Fix inconsistent core mesh editing hotkeys

Aleph0x:
This is a very small PR. I only changed two shortcut strings in the EdgeTool to make the hotkeys a bit more consistent with the rest of the mesh editor and with how users expect these operations to behave.

Changes
Bevel (edges): Alt+F -> F
Matches vertex bevel

Bridge Edges: Alt+B -> B
Move Mode -> t
Rotate Mode -> r
Scale Mode -> e
Pivot Mode -> d
All other core mesh editing operations use single key presses.

There is an ALT+B keybinding in hammer for BridgeTool Mode, but that creates an arch.

### DIFF
--- a/game/addons/tools/Code/Scene/Mesh/Tools/EdgeTool.UI.cs
+++ b/game/addons/tools/Code/Scene/Mesh/Tools/EdgeTool.UI.cs
@@ -147,7 +147,7 @@ partial class EdgeTool
 			return _edges.Length != 0;
 		}
 
-		[Shortcut( "mesh.edge-bevel", "ALT+F", typeof( SceneViewportWidget ) )]
+		[Shortcut( "mesh.edge-bevel", "F", typeof( SceneViewportWidget ) )]
 		private void Bevel()
 		{
 			if ( !CanBevel() )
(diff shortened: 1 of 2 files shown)


handsomematt:
Is this what you've already done @CarsonKompon ?

CarsonKompon:
No my changes fixed shortcuts not running with certain widgets in focus, this is just changing mapping tool default keybinds to copy hammers keybinds

CarsonKompon:
FWIW I’m happy to accept these changes. I have no real opinion on what they should be, just unsure if @aylaylay chose the existing keybinds for a reason (or has any other objections)

aylaylay:
The keybinds for move modes just match the ones we have in object tool. If we're going to decide to change them here, they should be changed in object tool too. Garry decided these keybinds.

I'm not sure why Bevel and Bridge have ALT, the only reason I can think of is that at the time they conflicted with something else so ALT was added.

Aleph0x:
After spending time with the new editor I think leaving these as they are now is fine

```
Bridge Edges: Alt+B -> B
Move Mode -> t
Rotate Mode -> r
Scale Mode -> e
Pivot Mode -> d
```

If anything we can just edit these hotkeys in the project setting to our liking. Drakefruit told me the original hotkeys probably have something to do with Unity.

I do think the inconsistent bevel hotkey between Edge mode and Vertex mode should be fixed. Bevel is used constantly, and having two different bindings for the same action is a bad experience out of the box. It should be a single, easy-to-press key. F is great since it’s so easy to reach, but if that's not possible due to conflicts then maybe another one should be decided?

PR #102, opened 2025-12-08
Show volume tooltip in settings as a percentage rather than a decimal

DrakeFruit:
Also increases the max volume from 100% to 200% because 100% is quieter than everything else on my pc and it's nice to have

### DIFF
--- a/engine/Sandbox.Engine/Game/Preferences.cs
+++ b/engine/Sandbox.Engine/Game/Preferences.cs
@@ -15,13 +15,13 @@ public static class Preferences
 	/// <summary>
 	/// The user's preferred Music volume, as set in the options, clamped between 0 and 1
 	/// </summary>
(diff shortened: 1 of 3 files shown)


garrynewman:
You make a nice PR that make sense, and then you change volume from 0-1 to 0-2

DrakeFruit:
200% volume is something I wish a lot more programs had, 100% is the default that everyone targets but what happens when I need it louder than that?

I could turn up my system volume but then everything else is too loud and I have to go through 5 programs and adjust the volume just to play s&box and every new program I install would be blasting until I turn it down.

I could remove it from the settings menu but a higher limit via console is much appreciated.

garrynewman:
I don't really know why there is a limit on the console command

DrakeFruit:
@garrynewman I removed the max part of the volume convars and set the max to 1 in the in game settings again. Thanks for the feedback

PR #105, opened 2025-12-08
RequestAsync support basic auth

766974616c79:
Using the above code, the builder returns a URL that is not supported by the HTTP client. You must use the Authorisation header to be able to use authentication.

Resolves https://github.com/Facepunch/sbox-public/issues/4306


766974616c79:
@garrynewman i made the recommended changes, sorry for the time taken

handsomematt:
Send the header manually if you want to use basic auth, the same way you would with a normal .NET HttpClient

PR #108, opened 2025-12-10
Prevent `INetworkListener.OnDisconnected` being called on clients

andy013:
Issue: https://github.com/Facepunch/sbox-public/issues/2372

### Changes
- Wrapped the calls to `OnDisconnected` in a `IsHost` if block. 

### Notes

The XML comment says this should only be called for the host, but it could be called for the client if the host disconnected. It's a bit tricky to reproduce since you can't use the editor or a dedicated server. I had to use 2 steam accounts, disable host transfer and then disconnect the host to reproduce it.


kurozael:
Yeah we should probably have this guard, but in this particular case wouldn't the client just be booted to menu?

andy013:
Yes, but I think it can happen when the host transfers as well. It just happens for the people who stay as clients. I've not confirmed this though as I was only testing with 2 accounts. When I tried that the other client always became the host before `OnDisconnected` was called.

PR #109, opened 2025-12-10
Implement audio event processors

Ardivaba:
Adds per-sound audio processors that can be applied to individual sound handles before mixing.

```c#
var processor = new HighPassEventProcessor();
var handle = Sound.Play( SoundEvent, WorldPosition );
handle.AddProcessor( processor );
```

**Why:**
Currently, AudioProcessor only works at the mixer level, processing all mixed sounds together. This makes it
impossible to apply effects to individual sound sources without creating a separate mixer for each sound event.
Per-sound processors enable use cases like sampling the scene around an audio source to create dynamic effects
(e.g., occlusion, reverb based on environment).

**Example:** https://www.youtube.com/watch?v=J5kPTVlQRBQ

**Changes:**
- New AudioEventProcessor base class for processing samples on individual sounds
- Built-in example processor: HighPass
- Processor chain management on SoundHandle


Ardivaba:
Removed WIP flag as implementing a component to manage processors would be trivial and in many cases makes sense to leave to developer discretion.

Ardivaba:
Failed test seems unrelated.

SwagAccount:
yes please

garrynewman:
There is something here but I feel like we need to be a lot more considered with our approach.

ardivaba-wenture:
> There is something here but I feel like we need to be a lot more considered with our approach.

I agree that the user facing API could be better but I'm not sure how to proceed here. I could propose various solutions in the PR while at the same time making this PR unusable.

The way I'd like to proceed is to continue to build examples for this use-case and let you guys figure out the correct implementation unless there's something specific I could work on.

PR #110, opened 2025-12-10
Add Collision Rules to NavMeshAgent Ground Trace

SwagAccount:
Changes The FindPhysicsGroundZ function on NavMeshAgentSystem so it includes the collision rules of the agent.


SwagAccount:
FFS, i didn't need to add it 3 times

PR #115, opened 2025-12-11
[TerrainEditorTool] Replace Labels with LineEdits for size and opacity

AustinDotCodes:
Replacing the Labels with LineEdits for size and opacity in the TerrainEditorTool class. Leveraging LineEdits TextChanged action to update the slider.

Resolves issue [#9703](https://github.com/Facepunch/sbox-public/issues/423)


aylaylay:
This isn't how we go about it. We should be using control widgets.

PR #124, opened 2025-12-12
Sound preview misc fixes

boxrocket6803:
Sound preview playback previously always started at 0, now it starts at the playhead position.

https://github.com/user-attachments/assets/93caa3f9-9b90-49da-88d2-45b297059b07

Also added a check to only increment time while actually playing, previously upon reaching the end of a file with looping disabled it would reset the playhead position then add an extra RealTime.SmoothDelta. When playing the file again the start would be cut off slightly. Now it correctly parks itself at 0.


boxrocket6803:
Was gonna just fix the one thing but kept seeing other stuff and figured it'd be more convenient for all involved to do a larger pr instead of a bunch of smaller ones.

before

https://github.com/user-attachments/assets/e20c1311-efb1-429e-8038-274771970555

after

https://github.com/user-attachments/assets/b13133c5-5384-437e-bc0d-c371e42b805a

PR #128, opened 2025-12-14
Fix crash under Proton due to case-sensitivity

Kaydax:
For some God forsaken reason, forcing the path to be lowercase here under Proton causes a `System.IO.DirectoryNotFoundException` due to case-sensitivity. Removing `ToLowerInvariant()` in the creation of the `LocalFileSystem` fixes the issue and allows the game to boot under Proton again.

### DIFF
--- a/engine/Sandbox.Filesystem/LocalFileSystem.cs
+++ b/engine/Sandbox.Filesystem/LocalFileSystem.cs
@@ -11,7 +11,7 @@ internal class LocalFileSystem : BaseFileSystem
 	{
 		Physical = new Zio.FileSystems.PhysicalFileSystem();
 
-		var rootPath = Physical.ConvertPathFromInternal( rootFolder.ToLowerInvariant() );
+		var rootPath = Physical.ConvertPathFromInternal( rootFolder );
 		system = new Zio.FileSystems.SubFileSystem( Physical, rootPath );
 
 		if ( makereadonly )


garrynewman:
This should really be fixed in Proton, not here.

Kaydax:
> This should really be fixed in Proton, not here.

~~This is a regression with s&box itself, not Proton. There isn't an easy way to properly fix it in Proton due to how Linux is heavily case-sensitive. A better solution is to not assume case insensitivity in the code base, which I have another branch for that~~

Zipdox2:
> This should really be fixed in Proton, not here.

It should not. Changing case sensitivity in the entirety of Proton is very extreme. I would like to ask what problem using ToLowerInvariant fixes in the first place.

handsomematt:
image

Here's when I changed it 6.5 months ago, not a few days ago when the regression started - it was done because every other bit of filesystem code normalizes to lowercase, so you would be removing it in one place only and fucking up everything it fixed before that.

I don't want to accept any fix that doesn't understand the problem fully, and doesn't understand the issues they'd cause throughout the rest of the engine.

Kaydax:
I apologize, the mistake was on my end. The entire issue was caused by a bug with steam that caused library paths with Uppercase paths to become all lower case, thus causing the problem. The fix is to remove the library and re-add it, fixing the issue.

PR #132, opened 2025-12-16
Fix UI not being resized to screenshot_highres dimensions on capture

trende2001:
- `screenshot_highres` now temporarily resizes Screen.Size to the requested output dimensions and flushes render targets
to avoid stale buffers.
- Restores previous screen size/custom size after capture

### DIFF
--- a/engine/Sandbox.Engine/Systems/Render/Multimedia/ScreenshotService.cs
+++ b/engine/Sandbox.Engine/Systems/Render/Multimedia/ScreenshotService.cs
@@ -1,4 +1,5 @@
 ﻿using NativeEngine;
+using Sandbox.UI;
 using System.Collections.Concurrent;
 using System.IO;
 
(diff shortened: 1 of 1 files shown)


engine/Sandbox.Engine/Scene/Components/Camera/CameraComponent.cs
```diff
@@ -416,6 +416,7 @@ private void OnCameraRenderUI( SceneCamera camera )
 			if ( target != this ) continue;
 			if ( RenderExcludeTags.HasAny( c.GameObject.Tags ) ) continue;
 
+			c.EnsureLayoutForSize( camera.Size );
```
lolleko: It's fine that higheres screenshot can uses a couple of hacks to get the correct resolution perspective etc..., but they shouldn't leak into other classes, anyway we can avoid this?

lolleko (changes requested):

engine/Sandbox.Engine/Scene/Components/Camera/CameraComponent.cs
trende2001: I moved my hack into ScreenshotService now

joshuascript:
I noticed this problem when using highres and one or two high res screenshots causes crashes due to running out of memory.

PR #133, opened 2025-12-17
Fixed razor Texture property using object reference instead of string

TheDimin:
Razor properties seem to always be set as string, any type gets converted to this, while `SetPropertyObject` exists, it was not used before.

This seems to cause issues for example a Texture gets converted to string path, then supposedly gets loaded from the string path? anyhow that doesn't really work, the path doesn't seem valid and more importantly this doesn't work for run-time created textures.

### DIFF
--- a/engine/Sandbox.Engine/Systems/UI/Controls/Image.cs
+++ b/engine/Sandbox.Engine/Systems/UI/Controls/Image.cs
@@ -9,17 +9,7 @@
 		/// <summary>
 		/// The texture being displayed by this panel.
 		/// </summary>
-		[Property]
-		public Texture Texture
-		{
-			get;
-			set
-			{
-				if ( field == value ) return;
-				field = value;
-				YogaNode.MarkDirty();
-			}
-		}
+		public Texture Texture { get; set; }
 
 		public override bool HasContent => Texture != null;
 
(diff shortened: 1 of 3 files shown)


CarsonKompon:
> With these changes it will become possible to use texture objects directly in razor instead of using src attribute.

You can already do this with `<Image Texture=@MyTexture />`. I'm not sure what the difference is between that and what this PR is doing

TheDimin:
Aha your right that the Property was already exposed... MY bad

BUT the actual issue seems that texture objects gets converted to string, causing them to break.
I reworked the PR so that it only has the changes to fix the issue shown below.
(The texture is a cloud reference, but i was getting the same result with local textures)

Before:
sbox-dev_VbB5vfIf7u

After:

sbox-dev_Gk6ti7p2aM

This is my razor code: 
```
<div class="entry-image-container">
     <img class="entry-image" Texture="@IconTexture" />
</div>
```

aidencurtis:
Previously, if you used `<Image />` with the Texture property, it would work as expected, but `` used the same way would break, as indicated in the message above. I can attest to that - though not sure about how the change might impact things otherwise.

TheDimin:
> Previously, if you used `<Image />` with the Texture property, it would work as expected, but `` used the same way would break, as indicated in the message above. I can attest to that - though not sure about how the change might impact things otherwise.

With Image it always worked ? I thought it was just a synonym. guess not.

Curious why my changes didn't impact <Image .

Anyhow facepuch let me know if this change is desired, otherwise I will swap to using <Image

garrynewman:
Yeah Image would have always worked because that's the actual Panel's class., so it would have been finding it. I'm not sure of the repercussions of this change, but it seems unneeded to me.

PR #140, opened 2025-12-23
Add SQLite database support

Asphaltian:
This resolves https://github.com/Facepunch/sbox-public/issues/862.

Adds persistent SQLite storage via `Microsoft.Data.Sqlite`. Databases are stored in the game's data folder (_sv.db_ for server, _cl.db_ for client)

### New classes
`Sql` - Static API
`SqlDatabase` - SQLite connection wrapper
`SqlBuilder` - Fluent query builder for dynamic SQL construction
`SqlTest` - Unit tests

### Usage
```csharp
// DDL and writes use Execute()
Sql.Execute( "CREATE TABLE IF NOT EXISTS players ( steamid INTEGER PRIMARY KEY, name TEXT, kills INTEGER DEFAULT 0 )" );
Sql.Execute( "INSERT OR REPLACE INTO players (steamid, name, kills) VALUES (@id, @name, @kills)", 
    new { id = 76561198012345678, name = "Player", kills = 5 } );

// Reads use Query()
var rows = Sql.Query( "SELECT * FROM players WHERE kills > @min", new { min = 10 } );
foreach ( var row in rows )
    Log.Info( $"{row["name"]} has {row["kills"]} kills" );

// Single value
var count = Sql.QueryValue<int>( "SELECT COUNT(*) FROM players" );
var topPlayer = Sql.QueryValue<string>( "SELECT name FROM players ORDER BY kills DESC LIMIT 1" );

// Explicit database when you need both client and server storage
Sql.Server.Execute( "INSERT INTO server_data ..." );
Sql.Client.Execute( "INSERT INTO client_prefs ..." );

// Transactions - manual
Sql.Begin();
Sql.Execute( "UPDATE players SET kills = kills + 1 WHERE steamid = @id", new { id } );
Sql.Execute( "INSERT INTO kill_log ..." );
Sql.Commit(); // or Sql.Rollback();

// Transactions - automatic (commits on dispose, rolls back on exception)
using ( var tx = Sql.Transaction() )
{
    Sql.Execute( "..." );
    Sql.Execute( "..." );
}

// Fluent query builder for dynamic queries
var query = new SqlBuilder()
    .Select( "steamid", "name", "kills" )
    .From( "players" )
    .Where( "kills > @min", new { min = 0 } )
    .OrderBy( "kills", descending: true )
    .Limit( 10 );

var topPlayers = query.ExecuteQuery( Sql.Server );

// Helpers
if ( !Sql.TableExists( "players" ) ) { ... }
if ( !Sql.IndexExists( "players", "idx_name" ) ) { ... }
var safe = Sql.Escape( "O'Brien" ); // O''Brien
var cols = Sql.GetTableColumns( "players" ); // ["steamid", "name", "kills"]
```

### Error handling
Methods return `null` (Query), `-1` (Execute), or `default` (QueryValue) on failure. Check `Sql.LastError` for the error message.
```csharp
var result = Sql.Query( "SELECT * FROM nonexistent" );
if ( result == null )
    Log.Warning( $"Query failed: {Sql.LastError}" );
```


garrynewman:
I struggle with this. I struggle whether it's a good idea or not, whether we should be thinking more advanced. Should we be adding a layer above this, like entity framework, that means that the database can be sqlite, mssql, mysql? Should we be having people write raw sql queries?

Asphaltian:
> I struggle with this. I struggle whether it's a good idea or not, whether we should be thinking more advanced. Should we be adding a layer above this, like entity framework, that means that the database can be sqlite, mssql, mysql? Should we be having people write raw sql queries?

I wrote this at first because I wanted to enable people to do simple SQL queries like you would in GMod, but an entity framework would be an interesting idea to expand upon. I'm fine with being limited to SQLite though

jusvit:
I personally would prefer exposing LiteDB (or a similar API) to be more useful and beginner friendly than direct Sqlite.

e.g. 
image

Entity framework would be ideal with the migration support but it's a bunch to abstract and probably shouldn't be abstracted since itself is an abstraction.

joshuascript:
Great idea. Not sure if raw table management is there, but database management is 100% needed.

PR #153, opened 2025-12-29
Make GameObjectSource public

yuberee:
This is set on each child body of ModelPhysics to report collision events to the main component.
No reason to have it internal as we also need it for features such as car doors or other child objects without having to create a new component on each for the sole purpose of listening to collision events.


aylaylay:
Only modelphysics uses GameObjectSource so that collisions are reported to modelphysics and not their rigidbodies, there's no reason for anything else to set GameObjectSource so it will be staying internal

PR #159, opened 2026-01-07
Fix inconsistencies in path naming in repo.

MrSoup678:
This is not a full fix. The only parts fixed are the trivial ones.

Partially fixes #130 on `sbox-server` side.


MrSoup678:
`git mv` was too destructive here.

PR #175, opened 2026-01-14
Fix cached Gizmo.Draw SceneModels not moving

boxrocket6803:
fixes https://github.com/Facepunch/sbox-public/issues/317 (which happens to be closed, cant win em all)

Previously anything skinned would get stuck in place at whatever transform it was in on the first call to `Gizmo.Draw.Model()`. This calls FinishBoneUpdate if the transform has changed such that the model properly updates its position. Fixes ModelDropObject as well as a few other specific facepunch components and a whole bunch of my own components. Only downside is for the BaseChair component or components like it FinishBoneUpdate would get called twice on frames where the model has moved, really can't imagine that being a performance issue though.


aylaylay:
Explained in the original issue. If your model is animated, you want to tick it outside the gizmo, not compute transforms twice.

PR #176, opened 2026-01-15
Make slnx / csproj always use relative paths

Kaydax:
Currently S&box will always generate solution files and csproj files with absolute paths. This causes issues for those who are running the editor under proton, which will generate with absolute paths according to proton. That makes it annoying to work on projects using Linux native tools as the paths are specific to wine and not the host system.

What this changes is making both the solution files and csproj files generate with relative paths if possible. This is platform agnostic and will not cause any issues under windows. This also handles the case as to where if the game and project files are on separate drives, it will default back to using absolute paths on windows.

Any feedback is appreciated

Fixes https://github.com/Facepunch/sbox-public/issues/4383 and https://github.com/Facepunch/sbox-public/issues/6512

### DIFF
--- a/engine/Sandbox.SolutionGenerator/Generator.cs
+++ b/engine/Sandbox.SolutionGenerator/Generator.cs
@@ -133,8 +133,8 @@ namespace Sandbox.SolutionGenerator
 				{
 					ProjectName = p.Name,
 					ProjectReferences = "",
-					ManagedRoot = AttemptAbsoluteToRelative( managedFolder, p.CsprojPath ),
-					GameRoot = AttemptAbsoluteToRelative( relativePath, p.CsprojPath ),
+					ManagedRoot = AttemptAbsoluteToRelative( p.CsprojPath, managedFolder ),
+					GameRoot = AttemptAbsoluteToRelative( p.CsprojPath, relativePath ),
 					References = p.References,
 					GlobalStatic = p.GlobalStatic,
 					GlobalUsing = p.GlobalUsing,
--- a/engine/Sandbox.SolutionGenerator/Generator.cs
+++ b/engine/Sandbox.SolutionGenerator/Generator.cs
@@ -2,8 +2,8 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
-using System.Runtime.InteropServices;
 using System.Text.Json;
+using Sandbox.Utility;
 
 namespace Sandbox.SolutionGenerator
 {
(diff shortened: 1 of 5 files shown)


handsomematt:
What happens if my project is on a different drive e.g s&box installed on C: and project on D:

kaikenny:
Amazing

handsomematt (changes requested):
Thanks for the contribution.

I'm happy to make things easier on Linux as long as it doesn't cause problems for Windows.

I'm happy with:
* Standardizing path seperators to `/` from `\`
* Switching to relative paths 

However there's a lot of introduced complexity and unoptimized code in FixPathCasing, what's the goal of it, why does it need to enumerate the filesystem.

Kaydax:
> Thanks for the contribution.
> 
> I'm happy to make things easier on Linux as long as it doesn't cause problems for Windows.
> 
> I'm happy with:
> 
> * Standardizing path seperators to `/` from `\`
> * Switching to relative paths
> 
> However there's a lot of introduced complexity and unoptimized code in FixPathCasing, what's the goal of it, why does it need to enumerate the filesystem.

I think the goal was to walk the path to ensure proper capitalization, but realizing it now it makes no sense to even keep it considering its such a rare case where it would even make sense, and does nothing for windows. I'll try to remove it later today and simplify the entire logic

Kaydax:
Turns out the only way to really get the proper path casing is to walk the path. This really is only important on case-sensitive file systems as there is no real direct API to grab the proper canonical path. I did optimize the code and simplify it so hopefully its good enough. If you need anything more please let me know

Kaydax:
Making this a draft for right now to rework the entire PR

Kaydax:
Ok code is way better. Instead of path walking we are instead using GetFinalPathNameByHandleW from the win32 api. Using this allows us to properly grab the canonical path name with little effort and keeps the code as simple as possible.

This does mean this PR does contain code that is windows only and is calling a native function, but since I don't have access to the native engine, there is no way to make this cleaner. It would be very easy to replace GetCanonicalPath with a native function and remove the code from managed

handsomematt (changes requested):
It's way better, but either:

1. Can FileInfo do the exact same thing but is built into C#
2. Can you formalize GetFinalPathNameByHandleW into a utility function somewhere. We'll want to ifdef stub it out for native Linux too.

Kaydax:
> It's way better, but either:
> 
> 1. Can FileInfo do the exact same thing but is built into C#
> 2. Can you formalize GetFinalPathNameByHandleW into a utility function somewhere. We'll want to ifdef stub it out for native Linux too.

FileInfo cannot do the same thing as it does not resolve symbiotic links nor does it return the canonical path.

And sure, I'll put GetFinalPathNameByHandleW into its own utility function and wrap it in an ifdef, I was just waiting on an answer as to what to actually do with it before I went ahead with keeping it in C#

Kaydax:
@handsomematt Should this be good?

engine/Sandbox.System/Utility/NativeFileSystem.cs
```diff
@@ -0,0 +1,77 @@
+using System;
+using System.IO;
+using System.Runtime.InteropServices;
+
+namespace Sandbox.Utility;
+
+/// <summary>
+/// Provides native file system operations with platform-specific implementations.
+/// </summary>
+public static partial class NativeFileSystem
```
handsomematt: Make this internal

engine/Sandbox.SolutionGenerator/Generator.cs
```diff
@@ -17,59 +18,50 @@ public ProjectInfo AddProject( string type, string packageIdent, string name, st
...
+		/// <summary>
+		/// Converts a path to be relative to a base path, always returning forward slashes.
+		/// </summary>
+		private string AttemptAbsoluteToRelative( string basePath, string targetPath )
 		{
-			string baseNormalized = NormalizePath( basePath );
-			string relativeNormalized = NormalizePath( relativePath );
-			string baseEnding = string.Empty;
+			string targetFileName = string.Empty;
+			string baseDir = NativeFileSystem.GetCanonicalPath( basePath );
```
handsomematt: Do we need to call GetCanonicalPath 4 times, or shall we just call it once on the output?

handsomematt (changes requested):

kaikenny:
matt what more do u want

PR #180, opened 2026-01-15
Update ModelDropObject gizmo model

boxrocket6803:
relating to https://github.com/Facepunch/sbox-public/pull/175

Skinned models would previously get stuck in place when dragging them into the viewport, now they don't. Fixes the only instance of a facepunch component using a potentially skinned model in Gizmo.Draw.Model() without updating it.


aylaylay:
👍

PR #181, opened 2026-01-16
gitignore compiled menu content

boxrocket6803:
Update gitignore to exclude the compiled menu assets that get downloaded by bootstrap, not sure if there's a reason those paths weren't already in gitignore. Seems easier than everyone having their own local gitignore floating around between branches.


handsomematt:
Not this simple sorry, it's on my mind but this file is controlled from our sync workflow

boxrocket6803:
yeah figured it was something like that

PR #183, opened 2026-01-17
Linux fixes runtime for sbox-server

joshuascript:
This implementation is the minimum required fix for sbox-server to run natively on Linux.

Caveats:
Both libskiasharp.so and harfbuzz.so needed to be moved to the bin/managed folder manually for this to work. Typically the Windows-only DLLs sit in this same directory, but those are generated in linuxsteamrt64 and are unreachable by s&box at the moment.

May need to update Linux to ffmpeg8 for this to work.

A workaround was introduced that extends the lifetime of the marshaled struct and prevents native use-after-free. Initializing _ClientInfo_ structs would cause a SIGSEGV error. When going over the gdb log, it appears the datafields became corrupted during its runtime. This issue may exist on Windows and was behaving differently. Debug traces can be provided upon request for the SIGSEGV error if needed. The _ClientInfo_ struct seems to only exist for the lifetime of the method call, and is dereferenced later.

This also cannot use System.Speech, it is Windows only, so it is stubbed out.

Let me know if this helps or if I am barking up the wrong tree. There are negative implications to System.Speech not being available on a Linux-only server system.

- Joshua


joshuascript:
Sent PR wrong.

joshuascript:
Not sure why every material file was included when I updated my fork I'll try to fix it.

PR #184, opened 2026-01-17
Fix native Linux runtime for sbox-server

joshuascript:
**Minimum changes for sbox-server to run natively**

**Result**
Screenshot from 2026-01-15 23-21-51

**Caveats**
**If a LibSkia Error is thrown:** Both libskiasharp.so and harfbuzz.so need to be moved to the bin/managed folder manually for this to work. Typically the Windows-only DLLs sit in this same directory, but on Linux they are generated in linuxsteamrt64 and are currently unreachable by the engine.

A workaround was introduced that extends the lifetime of the marshaled struct and prevents native use-after-free. Initializing ClientInfo structs would cause a SIGSEGV error. When going over the gdb log, it appears the datafields became corrupted during its runtime. This issue may exist on Windows and was behaving differently. Debug traces can be provided upon request for the SIGSEGV. The ClientInfo struct seems to only exist for the lifetime of the method call, and is dereferenced later. The workaround is not a long-term fix, but maybe it can provide insight on what is actually happening to a developer snazzier than me.

This also cannot use System.Speech, it is Windows only, so it is stubbed out for any Linux runtimes.

Let me know if this helps or if I am barking up the wrong tree. There are negative implications to System.Speech not being available on a Linux-only server system. Do we think Proton is better for servers?

** may need ffmpeg8 installed to work

- Joshua


DrakeFruit:
:pray:

joshuascript:
I tried to fix the formatting and it went nuts on me. You can see the most recent commit and get the actual changes. I'll try again on Monday when I'm off of work.

joshuascript:
PR updated to latest version

MrSoup678:
> If a LibSkia Error is thrown: Both libskiasharp.so and harfbuzz.so need to be moved to the bin/managed folder manually for this to work. Typically the Windows-only DLLs sit in this same directory, but on Linux they are generated in linuxsteamrt64 and are currently unreachable by the engine.



This part is fixed by DllImportResolvers, see #10101 for my implementation.

MrSoup678:
>A workaround was introduced that extends the lifetime of the marshaled struct and prevents native use-after-free. Initializing ClientInfo structs would cause a SIGSEGV error. When going over the gdb log, it appears the datafields became corrupted during its runtime. This issue may exist on Windows and was behaving differently. Debug traces can be provided upon request for the SIGSEGV. The ClientInfo struct seems to only exist for the lifetime of the method call, and is dereferenced later. The workaround is not a long-term fix, but maybe it can provide insight on what is actually happening to a developer snazzier than me.

That part too is fixed (partially): c0c6ed0, and #10105 for setters.

joshuascript:
@MrSoup678 That's fine it's just a draft

PR #188, opened 2026-01-18
Add custom texture filtering/address mode to line renderer

wheatleymf:
## Changes
- Allows artists to control texture filtering (point/bilinear/trilinear/anisotropic) on line and trail renderers
- Obsoletes "Clamp" in SceneTrailObject's `TrailTextureConfig` struct, use `FilterMode` and `TextureAddressMode` properties instead. Default config will set anisotropic filtering with "wrap" UV address mode.
- No changes in line.shader, it still uses bindless sampler state. It is assigned from SceneLineObject.cs in `SamplerState` property's setter
- This also adds support for changing these sampler settings on TrailRenderer, BeamEffect and ParticleTrailRenderer components. 
- One exception for BeamEffect is that allows changing only texture filtering. Clamp/wrap address mode is determined by existing logic in the component, where `Wrap` is the default option, and `Clamp` is set if "Travel" feature is enabled. This functionality is left intact because I didn't want to alter existing logic for it 

## Why 
Some other renderers (like decals) allow changing texture filtering already, so I figured that it would be beneficial for artists to have the same control for line renderer. Additionally, instead of being restricted to just "clamp" bool, you can set any address mode now. Since line renderer is used as a foundation for trail renderer and beam effects, they got support for changing texture filtering as well. 

image
image

### DIFF
--- a/engine/Sandbox.Engine/Scene/Components/Effects/BeamEffect.cs
+++ b/engine/Sandbox.Engine/Scene/Components/Effects/BeamEffect.cs
@@ -1,4 +1,5 @@
-﻿using System.Runtime.CompilerServices;
+﻿using Sandbox.Rendering;
+using System.Runtime.CompilerServices;
 
 namespace Sandbox;
 
@@ -101,6 +102,11 @@ public sealed class BeamEffect : Component, Component.ExecuteInEditor, Component
 	/// This is pretty much the same as TextureOffset - but it's seperate so you can use offset for offset, and scroll to scroll.
 	/// </summary>
 	[Property] public ParticleFloat TextureScroll { get; set; } = 0.0f;
+
+	/// <summary>
+	/// Controls texture filtering on this beam effect.
+	/// </summary>
+	[Property] public FilterMode FilterMode { get; set; } = FilterMode.Anisotropic;
 
 	/// <summary>
 	/// Color gradient of the beam over its lifetime. Defines how the color changes from birth to death.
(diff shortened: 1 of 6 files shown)


lolleko:
Thanks for looking into this, definitely an oversight.
I wonder if we should just expose the entirety of SamplerState directly on the lines/beam beam component rather than having all these individual properties that get converted into a SamplerState later.
Could we maintain existing logic if we do that? Are there any downsides to doing it?

wheatleymf:
Yes, I think this would work for Line and Trail renderers and would be a cleaner solution overall, but I am not sure about BeamEffect:

```csharp
		if ( TravelBetweenPoints )
		{
			var lerp = TravelLerp.Evaluate( lifeDelta, beam.Rand( 3289 ) );
			var length = beam.StartPosition.Distance( beam.EndPosition );
			var chunklength = lineRenderer.Texturing.UnitsPerTexture;

			var chunksPerLength = length / chunklength;
			chunksPerLength += 1;

			lineRenderer.Texturing = lineRenderer.Texturing with
			{
				TextureAddressMode = Rendering.TextureAddressMode.Clamp,
				Offset = 1 + (lerp * -chunksPerLength)
			};

		}

```

It has a very specific use-case for wrap/clamp address modes (it expects Clamp address mode when "Travel" feature is enabled, and I think whole travelling effect depends on it) so I thought that it should let you choose only texturing mode.

wheatleymf:
> I wonder if we should just expose the entirety of SamplerState directly on the lines/beam beam component 

I tried replacing individual settings to a single SamplerState property where it's appropriate in my PR, but it seems that editing any values in it from component is impossible. (with and without [InlineEditor] attribute)

I have double checked if any other components in engine expose SamplerState as a whole property, but it seems that everything else only exposes individual settings, (like filter mode) and then builds a new sampler state separately like I did. 

It seems that I can only keep using separate sampler settings as of now, @lolleko what would be the right move for me here?

lolleko:
Let's stick with individual properties for now then, can always change it in the future.

lolleko (approved):

PR #194, opened 2026-01-19
Fix for buttons with no gameobject

boxrocket6803:
https://github.com/Facepunch/sbox-public/commit/c8d27185483a9c9e3d5f242de750ca17a47bbb16 broke control sheet buttons that don't have an associated gameobject (e.g. adding a control sheet in an editor tool), this uses the previous `SceneEditorSession.Scope()` method if `GetContainingGameObject()` is null.


solwllms:
Thanks for the contribution, but there were a couple other places where this was an issue so I went for a more general fix 83ac6e0

PR #196, opened 2026-01-20
Fix cloud assets not persisting between sessions

trende2001:
- Clean up stale file entries from old package revisions to prevent validation failures on restart
- Fix ValidatePackage to only validate files matching the current package revision

This fixes Hammer cloud asset faulty persistence too, before some assets could go missing after restarting s&box.

### DIFF
--- a/engine/Sandbox.Tools/Assets/CloudAssetDirectory.cs
+++ b/engine/Sandbox.Tools/Assets/CloudAssetDirectory.cs
@@ -104,8 +104,6 @@ internal class CloudAssetDirectory : IDisposable
 
 		packages.Insert( package );
 
-		// Force checkpoint to ensure data is persisted to disk
-		db.Checkpoint();
 
 		packageCache[package.FullIdent] = package;
 	}
--- a/engine/Sandbox.Tools/Assets/CloudAssetDirectory.cs
+++ b/engine/Sandbox.Tools/Assets/CloudAssetDirectory.cs
@@ -91,7 +91,21 @@ internal class CloudAssetDirectory : IDisposable
 	public void AddPackage( Package package )
 	{
 		packages.DeleteMany( x => x.FullIdent == package.FullIdent );
+
+		// Clean up file entries from old revisions to prevent validation failures on restart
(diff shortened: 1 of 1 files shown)


solwllms:
Thanks for taking a look at this!

Do we need to do `db.Checkpoint()`? Afaik that won't be the difference between written or not, just housekeeping on merging logged operations into the main db file. Mindful that might be added perf cost unnecessarily since we add packages in bulk on startup, etc. Not sure if that'll throw up issues when done concurrently either.

But cleaning up files makes sense for sure. We used to uninstall before installing, so likely a blind spot from that change.
Did you manage to get a decent repro? I struggled when looking at this before, so just curious.

trende2001:
> Thanks for taking a look at this!
> 
> Do we need to do `db.Checkpoint()`? Afaik that won't be the difference between written or not, just housekeeping on merging logged operations into the main db file. Mindful that might be added perf cost unnecessarily since we add packages in bulk on startup, etc. Not sure if that'll throw up issues when done concurrently either.
> 
> But cleaning up files makes sense for sure. We used to uninstall before installing, so likely a blind spot from that change. Did you manage to get a decent repro? I struggled when looking at this before, so just curious.

I realised we don't need db.Checkpoint() now it seems since it's unnecessary overhead. I did manage to get a decent repro (specifically with map addons and Hammer maps missing cloud materials/models that were there last session before restarting s&box)

I tested an old map addon project (that had an old revision of the parent game the addon was pulling from), and it seemed to pull the latest things without a struggle (before this fix it would sometimes keep old prefabs and not pull in newly added ones)

Map addon repro
1. Have a map addon referencing a game package (an old one!)
2. Package gets updated with new prefabs/content
3. Open the map addon project
4. Before new prefabs won't appear / aren't pulled in
5. After: Stale entries cleaned, new revision's content pulled correctly

Hammer map repro
1. Make a new map in Hammer
2. Make a mesh and give it a cloud material
3. Drag some cloud models in
4. Save
5. Restart s&box
6. Next time you open the map, those cloud assets should reliably be pulled.

Aside from these it should benefit the engine overall.

PR #197, opened 2026-01-21
Make hi-res screenshots use MSAA value set by user instead of hardcoded 16X

wheatleymf:
## Changes
Currently `screenshot_highres` will capture the image with hardcoded 16X multisampling. This is OK for some cases, but I think it would make more sense if highres screenshots would respect your current MSAA level from settings instead. 

## Why
Gives more control to user how exactly hi-res screenshots look like. 

This is coming from my experience making screenshots and other media for Rust projects, but usually high res screenshots are processed in any image editing software afterwards, and screenshots with no antialiasing are better suited for these tasks. (result is more sharper/smoother, depending on how you process it)


lolleko:
Nice thanks, i think we added highres screenshots before we had that setting.

PR #199, opened 2026-01-22
Fix Asset Inspector showing "2 Assets" when creating a new Asset

trende2001:
# Fix
- Fixes a bug where creating a new non-engine resource (like a GameResource) and then clicking on it would cause the Inspector to incorrectly show "2 Assets selected"

# How did I get to this hypothesis?

I started by tracing the selection flow from Asset Browser to the Inspector:
1. AssetList.OnSelectionChanged fires with selected items
2. AssetBrowser extracts Asset objects and calls OnAssetHighlight (single) or OnAssetsHighlight (multiple)
3. EditorUtility.InspectorObject is set
4. AssetInspector reads so.Targets to get the asset count

The Inspector showing "2 Assets" meant so.Targets had 2 items, which meant Selection contained 2 entries.
The hash code was being mutated. 
```cs
  public void Rename( string newName )
  {
      FileInfo.MoveTo( FileInfo.GetNewPath( newName ) );  // Changes FileInfo.FullName
      Asset = AssetSystem.RegisterFile( FileInfo.FullName );
  }
```

This changes the object's hash code because FileInfo.FullName was being changed, while it's a key in SelectionSystem's dictionary.

It was relying on old behavior from the Asset Browser unnecessarily clearing the AssetList when SetItems already handled it.

### DIFF
--- a/game/addons/tools/Code/Editor/AssetList/Entries/AssetEntry.cs
+++ b/game/addons/tools/Code/Editor/AssetList/Entries/AssetEntry.cs
@@ -8,6 +8,14 @@ public class AssetEntry : IAssetListEntry
 {
 	private readonly Color TypeColor;
 
+	/// <summary>
+	/// Cached path used for identity (Equals/GetHashCode). This is set once at construction
+	/// and never changes, even if the file is renamed. This is critical because AssetEntry
+	/// is used as a key in hash-based collections (like SelectionSystem), and changing the
+	/// hash code after insertion would corrupt the collection's state.
+	/// </summary>
+	private readonly string _identityPath;
+
 	public readonly FileInfo FileInfo;
 	public readonly string TypeName;
 	public Asset Asset { get; private set; }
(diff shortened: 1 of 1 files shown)


handsomematt:
What

wheatleymf:
I actually noticed that editor kept telling me that I have two decal resources selected in asset browser even though that wasn't true, wonder if this PR fixes the problem. Started happening after creating a new decal resource in asset browser, couldn't get rid of this without restarting.

trende2001:
> I actually noticed that editor kept telling me that I have two decal resources selected in asset browser even though that wasn't true, wonder if this PR fixes the problem. Started happening after creating a new decal resource in asset browser, couldn't get rid of this without restarting.

It does.

bryiac:
I've had this issue as well for shader resources

Fortune117:
I've been getting this when creating new sound events

DrakeFruit:
insightful comment as always @handsomematt 

this pr would be huge

handsomematt:
Making an AssetEntry pretend it's something else isn't the solution here

solwllms:
Cheers for having a shot at this, but the issue should be resolved with 32368cd

trende2001:
It is, I was about to close it, thank you @solwllms

PR #9899, opened 2026-01-26
Update sbox-issues links

nixxquality:
## Summary

I was just going to fix the `Report a bug` menu item in the Editor opening the old issue tracker but I figured I might as well update all the links while I was at it.

## Motivation & Context

Well the menu item should definitely be updated. The links on the comments are arguable since it dirties the git blame and they redirect correctly anyway (for now?).
Let me know and I'll be more picky with my changes.

## Implementation Details

For all the links in the comments I clicked them and see where I was redirected, then I copied that URL back in.

## Checklist


handsomematt (changes requested):
Thanks for the thorough changes. We can't currently accept changes to the readme files with how our sync is set up, could you remove those changes? (I've pushed something similar that will propagate soon)

PR #9902, opened 2026-01-26
Select the implementation without parameters

766974616c79:
## Summary

This PR eliminates errors if a structure has multiple implementations of ToString.

## Motivation & Context

The editor generates an error if there are multiple implementations of ToString.

```csharp
struct Test
{
	public override string ToString()
	{
		return "ekip";
	}
	
	public string ToString(int test)
	{
		return "ekip";
	}
}
```

Fixes: #222

## Implementation Details

Instead of using the `GetMethod` method, I use `GetMethods` and then search for ToString implementations and then implementations with no parameters.

## Checklist


handsomematt:
Good find, but why not this

image

766974616c79:
you're right, in trying to fix this problem, i looked at the flags but not at the various implementations of the function...

PR #9919, opened 2026-01-28
[DDGI] Stability Improvement for  ComputeVisibility():  Added Stabilizer for variance floor & Delta Clamp to avoid tails

GapingPixel:
Added Stabilizer for variance floor & Delta Clamp to avoid tails


sampavlovic:
Seems right, can't find any meaningful difference even in edge case tests though, could you give more details on the edge cases this is solving

DrakeFruit:
With the lack of pictures I feel like this is vibe coded and they just asked claude what's wrong with ddgi

sampavlovic:
Iterating on this and getting somewhere, I get what it's been trying to do now

ezgif-45af2699792aef40

PR #9920, opened 2026-01-28
Fix world input

766974616c79:
## Summary

Having the World Input component spams errors

## Motivation & Context

Stop having errors

Fixes: #168

## Implementation Details

My understanding of the problem is that `GetPanel` returns a null value because `worldPanel` is initialized when `OnEnabled` executes.

My solution is to check if the value of `GetPanel` is null.

## Checklist


766974616c79:
https://github.com/Facepunch/sbox-public/commit/eccaf21746683788a038dc94d6c82044e6a2873c

DevulTj:
Didn't actually notice this -- better to just isolate to Game.ActiveScene though imo.

PR #9934, opened 2026-01-29
Add BaseLength

yuberee:
This change adds the TargetLength property, which gives developers the ability to dynamically change the length of the rope instead of relying on the initial length calculated.

Example:
Reeling in a catch with your fishing rod, the rope needs to change length depending on how much string you've got left.

### DIFF
--- a/engine/Sandbox.Engine/Scene/Components/Game/VerletRope.cs
+++ b/engine/Sandbox.Engine/Scene/Components/Game/VerletRope.cs
@@ -21,10 +21,10 @@ public class VerletRope : Component, Component.ExecuteInEditor
 	public LineRenderer LinkedRenderer { get; set; }
 	
 	/// <summary>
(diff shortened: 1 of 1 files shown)


yuberee:
Updated:

- TargetLength -> BaseLength and better summary
- Use the new Advanced attribute instead of a group called "Advanced"
- BaseLength setter now doesn't allow a value under 0f (Similar to [this](https://github.com/Facepunch/sbox-public/blob/188a79df6e8b7554d1ec4626997c55f9e27711d2/engine/Sandbox.Engine/Utility/Mouse.cs#L40))
- Modifying BaseLength or Slack now set isAtRest to false to reenable simulation (Similar to [this](https://github.com/Facepunch/sbox-public/blob/188a79df6e8b7554d1ec4626997c55f9e27711d2/game/editor/ActionGraph/Code/WarningFrame.cs#L16))

PR #9951, opened 2026-02-02
Add NetList<T>.Slice(start, length)

nixxquality:
## Summary

This PR adds the Slice function to NetList&lt;T&gt;. List&lt;T&gt; has it as well: https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.slice?view=net-9.0

## Motivation & Context

I don't remember exactly why I needed this at the time, but it bothered me enough to submit an issue at one point.

Fixes: #1343

## Implementation Details

This is the core of the function:
```cs
		var output = new List<T>( length - start );
		for ( var i = start; i < start + length; i++ )
		{
			output.Add( _list[i] );
		}
		return output;
```
I'm not sure this is the best way, but I did the basics of avoiding LINQ since apparently it does bad allocations, and initializing the output list with a known capacity.
I assume the compiler knows what to do with this to make it better.

For reference, this is how List&lt;T&gt;.Slice is implemented:
```cs
            List<T> list = new List<T>(count);
            Array.Copy(_items, index, list._items, 0, count);
            list._size = count;
            return list;
```
https://github.com/dotnet/runtime/blob/1d1bf92fcf43aa6981804dc53c5174445069c9e4/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/List.cs#L690-L693

Since I don't have access to the ObservableCollection&lt;T&gt; internals, and since that doesn't have a Slice function, that's why I decided to make a simple for loop.

## Checklist


handsomematt:

PR #9962, opened 2026-02-03
Async AcceptConnection

nixxquality:
## Summary

This PR replaces AcceptConnection with an async version. The original functionality is left in place, but marked as Obsolete.

## Motivation & Context

An async function lets you make a time-consuming call like a check with a database or on-disk ban list without causing a server freeze on connections.

Fixes: #1563

## Implementation Details

Discussion happened in Discord: https://discord.com/channels/833983068468936704/895597253316714497/1468245963356508322
The initial implementation was just a string where null means no rejection.
Further discussion led to a new ConnectionRequestResult struct with static defaults so it can be expanded in the future if need be.

**Note**: I have no idea where the ConnectionRequestResult struct should go. I can either follow instructions or you can feel free to pull in the PR and edit it to your hearts content.

(Note: this video was recorded before the rework but not much has changed)

https://github.com/user-attachments/assets/585d87e7-369b-43cc-9b02-e574ca9b2447

Sample projects:
- [rejectconnection initial.zip](https://github.com/user-attachments/files/25050006/rejectconnection.initial.zip)
- **[rejectconnection struct.zip](https://github.com/user-attachments/files/25050012/rejectconnection.struct.zip)**

## Checklist


badandbest:
I'd recommend squashing the commits into one when you're done

engine/Sandbox.Engine/Scene/Networking/SceneNetworkSystem.cs
```diff
@@ -644,13 +644,27 @@ public override bool AcceptConnection( Connection channel, ref string reason )
...
 		return true;
 	}
 
+	public override async Task<ConnectionRequestResult> AcceptConnection( Connection channel )
+	{
+		foreach ( var c in Game.ActiveScene.GetAll<Component.INetworkListener>() )
+		{
+			var result = await c.AcceptConnection( channel );
+
+			if ( !result ) return result;
```
badandbest, line 663: You could combine both `SceneNetworkingSystem.AcceptConnection` methods by making the obsolete `INetworkListener.AcceptConnection` return a `ConnectionRequestResult.Reject`. This would simplify other parts of the class.

```suggestion
#pragma warning disable CS0618 // Games still use the old method.
			var denialReason = "";

			if ( !c.AcceptConnection( channel, ref denialReason ) )
				return ConnectionRequestResult.Reject( denialReason );
#pragma warning restore CS0618

			var result = await c.AcceptConnection( channel );

			if ( !result ) return result;
```

nixxquality: The double functions exist because `AcceptConnection` is a public member of the `public abstract partial class GameNetworkSystem`. Changing that would be a breaking change.
https://github.com/Facepunch/sbox-public/blob/69a2985c8645c3b94a70175596e8f86f78e91706/engine/Sandbox.Engine/Systems/Networking/System/GameNetworkSystem/GameNetworkSystem.cs

badandbest: Is it even possible for games to get access to GameNetworkSystem?

nixxquality: Not sure what you'd use it for, but again, it is public so yes it can be accessed.

nixxquality:
As an aside, I noticed that if an exception is thrown in the AcceptConnection function (both current and new) then the connecting user gets stuck in the loading screen until he gives up and presses cancel.

I could fix that in this PR as well, if you would like me to.

badandbest:
> if an exception is thrown in the AcceptConnection function then the connecting user gets stuck in the loading screen until he gives up and presses cancel

Yeah you should probably fix that right away

trende2001:
This'd be great

handsomematt:

PR #9983, opened 2026-02-06
Add fade-in support for Sound.Play

matekdev:
# Pull Request

---

## Summary

This PR adds in the ability to fade-in a sound. This is similar to the functionality that is present in `SoundHandle.Stop(...)` where you can optionally pass in a fade-out time.

## Motivation & Context

Fixes: 

## Implementation Details

Follows the exact same structure as `fade-out`, nothing crazy.  In the case where we are fading-in and fading-out, I made the choice of fading-out taking precedence over `fade-in`. 

My only two concerns are...
1. Are we okay with passing an extra parameter like this into `Sound.Play`? We do for `SoundHandle.Stop(...)`
2. Let's make sure we don't break any current APIs

## Checklist

### DIFF
--- a/engine/Sandbox.Engine/Systems/Audio/Sound.Play.cs
+++ b/engine/Sandbox.Engine/Systems/Audio/Sound.Play.cs
@@ -7,19 +7,19 @@ public static unsafe partial class Sound
 	{
 	}
 
-	public static SoundHandle Play( string eventName )
+	public static SoundHandle Play( string eventName, float fadeInTime = 0.0f )
 	{
 		if ( Application.IsHeadless )
 			return SoundHandle.Empty;
 
 		var e = SoundEvent.Find( eventName );
-		if ( e is not null ) return Play( e );
+		if ( e is not null ) return Play( e, fadeInTime );
 
 		Log.Info( $"Couldn't find sound event {eventName}" );
 		return default;
 	}
 
-	public static SoundHandle Play( SoundEvent soundEvent )
+	public static SoundHandle Play( SoundEvent soundEvent, float fadeInTime = 0.0f )
 	{
 		if ( Application.IsHeadless )
 			return SoundHandle.Empty;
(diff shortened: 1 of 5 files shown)


engine/Sandbox.Engine/Systems/Audio/Sound.Play.cs
```diff
@@ -31,7 +31,7 @@ public static SoundHandle Play( SoundEvent soundEvent )
 		if ( !soundFile.IsValid() )
 			return null;
 
-		var handle = PlayFile( soundFile, soundEvent.Volume.GetValue(), soundEvent.Pitch.GetValue() );
+		var handle = PlayFile( soundFile.native.self, soundEvent.Volume.GetValue(), soundEvent.Pitch.GetValue(), 0.0f, fadeInTime );
```
lolleko: why does this use the native overload?

should probably use
`public static SoundHandle PlayFile( SoundFile soundFile, float volume = 1.0f, float pitch = 1.0f, float delay = 0.0f, float fadeInTime = 0.0f )`

so we get the debug name set.

boxrocket6803:
if it starts fading out before its fully faded in it looks like it snaps to full volume, maybe instead of else if it should just fade in/out at the same time? might sound smoother

matekdev:
> if it starts fading out before its fully faded in it looks like it snaps to full volume, maybe instead of else if it should just fade in/out at the same time? might sound smoother

It would take the current value of `volume` and then multiply whatever `voice.Fadeout.EvaluateDelta(...)` evaluates to. 

Happy to do whatever approach Facepunch would like.

lolleko:
So the issue is fade in and fade out overlapping?

I think which to pick comes down to personal preference.
As discontinuities are often quite audible i would personally lean towards blending rather than overriding with fade out.

Idk know how feasible but if possible it would be nice to have a little validation in editor using the validate property attribute. That throws up a warning if they overlapp.

PR #9989, opened 2026-02-06
Blender Style Rotation Gizmos

DrakeFruit:
## Summary

This PR adds a trackball and view-space rotation gizmo, similar to how blender does it

## Motivation & Context

I'd like to rotate multiple axis at once, or rotate something in view-space. This is really useful for mapping in general and is part of getting feature parity with hammer and other 3D applications.

Fixes: 

## Implementation Details

Unfortunately to get this working I had to swap the output type of gizmocontrols.rotate to a rotation instead of angles. This is a breaking change for anything that uses this method but I have gone ahead and adjusted any engine code that uses this to reflect the change.

I've also added a gizmo method for a distance vector, the only existing method returned a float.

[Screencast_20260206_125638.webm](https://github.com/user-attachments/assets/c530b2a6-c1aa-4454-a2ea-add9e3a66e4a)

## Checklist

- [x] Fix world vs local space rotation
- [x] Fix trackball "drifting" when continuously rotating, should use initial grab point and delta from that point instead of current delta
- [x] Potentially add blender / hammer style rotation indicators with a pie circle that gets filled
- [ ] Unit tests?

### DIFF
--- a/engine/Sandbox.Engine/Editor/Gizmos/Control/Control.BoundingBox.cs
+++ b/engine/Sandbox.Engine/Editor/Gizmos/Control/Control.BoundingBox.cs
@@ -7,7 +7,7 @@ public static partial class Gizmo
 		private static bool ArrowPoint( string name, Vector3 direction, float length, Color color, out float distance, ref bool pressed )
 		{
 			distance = 0.0f;
-			var rotation = Rotation.LookAt( direction, Vector3.Up );
+			var rotation = global::Rotation.LookAt( direction, Vector3.Up );
 			using var x = Scope( name, direction * length, rotation );
 			var worldBoxScale = Transform.UniformScale;
 			using var scaler = PushFixedScale();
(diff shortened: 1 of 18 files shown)


DrakeFruit:
https://github.com/user-attachments/assets/a2d1038f-9e1c-44f4-955c-16655653668d

fixed the inconsistency, might need to multiply rotation speed by 1.5 or something, feels a little slow

DrakeFruit:
API change seems to break my summer cottage, anyone have some advice on this? I don't think it's possible to not break the API here

DrakeFruit:
API change seems to break my summer cottage, anyone have some advice on this? I don't think it's possible to not break the API here @bakscratch

bakscratch:
> API change seems to break my summer cottage, anyone have some advice on this? I don't think it's possible to not break the API here @bakscratch

I don't really know what any of that means, but you also removed the snapping too? Just looking at the code

handsomematt (changes requested):
This is good. But it's an API breaking change because you're changing

```diff
- public bool Rotate( string name, out Angles outValue )
+ public bool Rotate( string name, out Rotation outValue )
```

It seems like adding an Angles overload that uses the new function with .ToRotation() would work fine, and marked as obsolete.

DrakeFruit:
okay should be all fixed now, added back snapping too though the trackball snapping is a little strange because I'm only snapping the rotation around the axis and not snapping the rotation of that axis itself. I can't be asked to do this tonight maybe later. Or it's fine without

PR #10013, opened 2026-02-11
Preserve remote URL casing in texture lookup normalization

binbonn:
## Summary

This PR fixes a texture loading issue where `` URLs could be lowercased during texture path normalization.  
Some remote image paths are case-sensitive, so this could break valid image URLs.

## Motivation & Context

Remote URLs and data URIs should preserve their original casing.  
The previous normalization path applied lowercase rules intended for local file paths to all texture inputs.

Fixes: #9986 

## Implementation Details

- Added `NormalizeLookupPath` in `Texture.Load.cs` to centralize lookup normalization behavior.
- Kept existing lowercase normalization for local file paths.
- Preserved original casing for:
  - `http/https` URLs
  - `data:` URIs
- Applied the same normalization path in both:
  - `LoadInternal` (load/cache insert path)
  - `Find` (cache lookup path)

This keeps cache behavior consistent while preventing URL casing regressions.

N/A (non-visual fix)

## Checklist

- [N/A] Public APIs are documented (if applicable)


handsomematt (changes requested):
I get the need to do this, but this is a lot of added complication.

The first call to normalize the filepath seems redundent, because everywhere it's passed to that needs to use it.. already normalizes it again.

So a non normalized path can be passed to the loaders just fine in the first place.

PR #10022, opened 2026-02-13
Enforce FriendsOnly lobby privacy during handshake

trende2001:
## Summary

Adds server-side enforcement for `LobbyPrivacy.FriendsOnly` lobbies during the handshake phase. When a lobby is set to FriendsOnly, the host now validates the joining player's SteamId against its own Steam friend list and kicks non-friends with the message "This lobby is Friends Only."

Previously, non-friends could connect and remain in the lobby.
## Motivation & Context
Players expect that setting their lobby to "Friends Only" actually prevents non-friends from joining. This change closes that gap at the engine level so game code doesn't need to reimplement it.

## Implementation Details
- The check runs in `On_Handshake_ClientInfo` after the SteamId is resolved from the trusted `SteamLobbyConnection`, using `Friend.IsFriend` against the host's local friend list.
- The host always bypasses the check. Steam API failures fail closed (deny access).

## Checklist

### DIFF
--- a/engine/Sandbox.Engine/Systems/Networking/System/NetworkSystem.Handshake.cs
+++ b/engine/Sandbox.Engine/Systems/Networking/System/NetworkSystem.Handshake.cs
@@ -144,6 +144,24 @@ internal partial class NetworkSystem
 
 		Log.Info( $"{msg.DisplayName} [{msg.SteamId}] is connecting" );
 
+		//
+		// If the lobby is set to FriendsOnly,
+		// only allow players who are Steam friends with the host.
+		//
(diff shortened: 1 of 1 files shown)


Kaydax:
LGTM

handsomematt (changes requested):
Why is there a huge try catch around `isFriend = joiningFriend.IsFriend;`?

This is a good place to do this friend check, but your implementation can be a lot tighter.

trende2001:
> Why is there a huge try catch around `isFriend = joiningFriend.IsFriend;`?
> 
> This is a good place to do this friend check, but your implementation can be a lot tighter.

I made the implementation a lot more tighter without over defensive checks.

PR #10043, opened 2026-02-19
Allow single parameter in [Change] method

boxrocket6803:
## Summary

Adds support for a single parameter `OnChanged(T oldValue)` in addition to the existing `OnChanged()` and `OnChanged(T oldValue, T newValue)` formats for methods invoked by the Change attribute.

## Motivation & Context

The Change attribute allowed either a method with no parameters or a method with parameters for both the old and new values. Because the property is already set to the new value when the change method is called passing the new value into the method as a parameter seems a bit pointless.

image

## Checklist


badandbest:
I see where you're coming from, but if I saw a method called NewTarget or OnTargetChanged I'd expect it to receive the **new** value.

There's already precedent for passing the new value in the engine:
https://github.com/Facepunch/sbox-public/blob/29dfeaaf7abee63c0a88ca6d172f165ea886929d/engine/Sandbox.Engine/Systems/SceneSystem/SceneObject.cs#L129-L141

https://github.com/Facepunch/sbox-public/blob/29dfeaaf7abee63c0a88ca6d172f165ea886929d/engine/Sandbox.Engine/Game/PartyRoom/PartyRoom.cs#L174-L181

https://github.com/Facepunch/sbox-public/blob/e66a1a64e2fbfdd0b723db0053ac5228103d583d/engine/Sandbox.Tools/Qt/Widget.Events.cs#L443-L451

https://github.com/Facepunch/sbox-public/blob/e66a1a64e2fbfdd0b723db0053ac5228103d583d/game/addons/tools/Code/Editor/ExpandGroup.cs#L102-L125

engine/Sandbox.Engine/Systems/Console/ConsoleSystem.cs
```diff
@@ -43,6 +43,12 @@ public static void OnChangePropertySet<T>( in WrappedPropertySet<T> p )
 			x.Parameters[0].ParameterType == typeof( T ) &&
 			x.Parameters[1].ParameterType == typeof( T ) );
```
badandbest: You could probably simplify this section by returning the first method whose parameters are all of type T.

```suggestion
			x.Parameters.All( p => p.ParameterType == typeof( T ) ) );
```

Then switch based on the parameter count.

```c#
switch ( method?.Parameters.Length )
{
	case 2:
		method.Invoke( p.Object, [oldValue, p.Value] );
		break;
	case 1:
		method.Invoke( p.Object, [oldValue] );
		break;
	case 0:
		method.Invoke( p.Object );
		break;
	default:
		Log.Warning( $"{type.Name}.{property.Name} has [Change] but we can not find {functionName}( {property.PropertyType} oldValue, {property.PropertyType} newValue )" );
		break;
}
```

boxrocket6803:
Seems kinda pointless to me to pass the new value in at all (unless the getter/setter does something weird but that seems rare) and I feel like keeping the order consistent makes more sense but I can also see how it might be confusing. Personally I always use the 0 parameter version unless I need the old value, and then I get annoyed that I cant use `old` and `new` for the parameter names. Perhaps simply explaining it in the summary for [Change] would do the trick?

lolleko:
I also don't really see the problem this solves.

Am I missing something or is the only reason we would want this, that we are too lazy to write `fun( X old, X _)` ?

boxrocket6803:
would indeed be the only reason, just looks a bit messy with the unused parameter

badandbest:
> I also don't really see the problem this solves.
> 
> Am I missing something or is the only reason we would want this, that we are too lazy to write `fun( X old, X _)` ?

You're not missing anything.

Being too lazy to write `fun( X old, X new )` was the same reason someone supported the parameterless version to the engine.

Being too lazy to write
```csharp
set
{
    if ( field == value ) return;
    field = value;
    OnChanged( value )
}
```
is why the ChangeAttribute exists in the first place. All of this is for convenience.

While I disagree with how it was implemented, I don't think this PR is unreasonable.

handsomematt:

PR #10046, opened 2026-02-19
Fix scene thumbnail regeneration causing map reload on every save

Ardivaba:
## Summary
- Fix `AssetSystem.Tick()` eagerly rebuilding thumbnails on every asset change by switching to lazy invalidation (`RebuildThumbnail(false)`), so thumbnails only regenerate when next viewed in the Asset Browser
- Fix `ListView.LayoutItems()` not firing `ItemScrollExit` callbacks when the items list becomes empty (e.g. during folder change), causing `EditorEvent.Unregister` to never be called — leaking event listeners that keep rebuilding thumbnails for assets the user is no longer viewing

Fixes #1344


handsomematt:

PR #10047, opened 2026-02-19
Fix GameResource with texture generator falsely marked as dirty on open

Ardivaba:
Fixes: #285

## Summary
- Split initial texture generation into a dedicated `InitialGenerate()` method that only updates the display (texture preview) without calling `SerializedProperty.SetValue()`, preventing false dirty state
- Gate `FrameUpdate()` behind an `_initialized` flag so any spurious dirty state from widget construction during initialization doesn't trigger a second `UpdateGenerator()` with `SetValue`
- Move `OnPropertyChanged` subscription in `AssetInspector` to after `sheet.AddObject()`, so synchronous property changes during control widget construction don't trigger `StateHasChanged()`

Opening a GameResource that uses a custom texture generator (e.g. decal definitions) would falsely mark it as having unsaved changes, causing a save dialog to appear when switching to another asset.

**Before**
https://github.com/user-attachments/assets/b78cc2f7-befb-4980-aa25-3c89ce59c820
**After**
https://github.com/user-attachments/assets/8f6ec394-d5f4-4262-bdb9-49353590d853


Ardivaba:
~~There's still issues with this, working on fixing them, hence the BROKEN: prefix.~~
Nvm, issue was unrelated.

handsomematt:

PR #10051, opened 2026-02-20
WIP: Add sounds to NS2 mount

nixxquality:
## Summary

This PR adds the missing sound files to the NS2 mount.

## Motivation & Context

I would like to use the NS2 sounds together with other NS2 assets.

## Implementation Details

The sound files in NS2 are grouped together in FMOD Sound Banks.
I based my implementation on https://github.com/HearthSim/python-fsb5

NS2 only uses MPEG and PCM16 formats inside the banks.

When NS2 is mounted, it will read the header of every bank and add a loader to the context for each sample inside.
The loader will then read inside the bank file at the correct offsets.

What makes this a WIP is that I have no idea how to create sounds from just samples and settings. I'm pretty sure that's internal-only for now. Therefore, I have two approaches for the actual 'loading' part:

1. For PCM16, I create a fake .wav file in memory and load it.
2. For MPEG files I write them to disk as a proof of concept.

Another note: NS2 makes liberal use of the LOOP metadata in the sample chunks. That information is just being skipped for now.

https://github.com/user-attachments/assets/55aa581b-b25a-4cb1-bc37-550bdb6fd206

## Checklist


handsomematt:

PR #10057, opened 2026-02-21
`display: inline-flex` support

### LINKED ISSUE #4803: UI: Support inline-flex
### For?

S&Box

### What can't you do?

Currently the only display values we have are `flex` and `none`. It would be useful to have `inline`, `block` and `inline-block`.
A use case I have is where I'm trying to use custom emojis inside my block of text.

`display: inline`, `display: block` and `display: inline-block`. Any other values I do not care about. But it might be good to ask what other people think.

### What have you tried?

When I have a line of text with a custom emoji/symbol, I've been setting the flex direction to column, making a div, and using several labels to keep the text and custom emoji on the same line.

### Additional context

_No response_
```[tasklist]
### Tasks
- [ ] display: inline-flex
```

trende2001:
## Summary

Adds `display: inline-flex` as a supported CSS display value in the UI layout engine.

Inline-flex containers use flex layout internally for their children but shrink to fit
their content width/height rather than stretching to fill the parent - the same
distinction the CSS spec makes between `block` and `inline` elements, applied to flex
containers.

## Motivation & Context

The primary use case is embedding custom emojis or icons inline within text content.
Previously the only way to approximate this was with nested column-direction flex
workarounds. `display: inline-flex` lets a panel sit inline in a flow while still
managing its own children with full flex rules.

Fixes: #4803
Related: #1423

## Implementation Details

Yoga has no native `inline-flex` concept, so the semantics are layered on top at the `Panel.Layout.cs` level. When `IsInlineFlex` is true, Yoga is passed `DisplayMode.Flex` - the distinction is enforced by overriding a set of defaults before the layout pass:

- `Width`/`Height` -> `Length.Auto` (unless explicitly set in the panel's own `Style`)
- `FlexGrow`/`FlexShrink` -> `0` (don't stretch into parent)
- `AlignSelf` -> `FlexStart` (don't cross-axis stretch)

Explicit values in the panel's stylesheet always win over these defaults.

https://github.com/user-attachments/assets/ac089db4-fee7-4c78-b763-d7e46ed1d306

image

## Checklist

### DIFF
--- a/engine/Sandbox.Engine/Systems/UI/Panel/Panel.Layout.cs
+++ b/engine/Sandbox.Engine/Systems/UI/Panel/Panel.Layout.cs
@@ -235,13 +235,49 @@ public partial class Panel
 		if ( ComputedStyle == null )
 			return;
 
-		YogaNode.Width = ComputedStyle.Width;
-		YogaNode.Height = ComputedStyle.Height;
+		// Handle width specially for inline-flex: default to auto (shrink-to-content)
+		// unless an explicit width is set via inline style OR CSS/SCSS stylesheet
+		var hasExplicitWidth = Style?.Width != null ||
+			(ComputedStyle.Width.HasValue && ComputedStyle.Width.Value.Unit != LengthUnit.Undefined);
+
+		if ( IsInlineFlex && !hasExplicitWidth )
+		{
+			YogaNode.Width = Length.Auto;
+		}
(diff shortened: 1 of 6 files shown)


handsomematt:
If all this is is a bunch of css properties, why does this need to exist? Setting display shouldn't just set a bunch of other properties.

PR #10061, opened 2026-02-21
Fix list reorder not updating inspector for class types

### LINKED ISSUE #10060: Reordering lists with classes does not update the inspector
### Version / Commit

26.02.18

### Describe the bug

https://github.com/user-attachments/assets/b8fc943c-2dc5-4752-bc72-bd14537ac765

Dragging the classes around does not update the inspector, structs seem to do it (sometimes?) but I'm not sure if that's just a side effect of them not being reference types.

I am quite certain this does not happen if it's a list of GameResources or Components or GameObjects.

### To Reproduce

Drag items around in a list of classes.

### Expected behavior

Inspector should update

### Media/Files

[list_update_repro.zip](https://github.com/user-attachments/files/25462017/list_update_repro.zip)

### Additional context

_No response_

Ardivaba:
Invalidate `buildHash` after a move operation so the inspector properly rebuilds when reordering list entries of class types.

  The `ValueHash` only considers the list reference and count, so reordering items doesn't change the hash. This causes `Rebuild()` to short-circuit, leaving the UI stale. Setting `buildHash = null` after a move forces the rebuild.

  Fixes #10060


Ardivaba:
Braxen made a great issue.
Every issue should require repro project and a video unless it's a feature request.

PR #10117, opened 2026-03-03
[Shadergraph] Change upgrade code to work similarly to how JsonUpgraders work

QuackCola:
## Summary

This Pull request tweaks the graph upgrade code to something that works pretty much exactly how existing JsonUpgraders work.

## Motivation & Context

This change makes it so upgrades to ShaderGraph assets in the future can be handled one after the other and make use of the same system that GameResources use to upgrade themselves when they need to.

## Implementation Details

Just define an upgrader similarly to how the other ones are defined like so :

```cs
[JsonUpgrader( typeof( ShaderGraph ), 1 )]
internal static void Upgrader_v1( JsonObject obj, JsonSerializerOptions options )
{
}
```

One thing of note is that the `JsonUpgrader` class is internal to the engine. Which means I had to make a simple copy of the class and called it `ShaderGraphJsonUpgrader` so that i could actually start the upgrade process. If the  `JsonUpgrader` class is made public it can be easily changed so i just use what's already there instead of having a near identical copy of the static class.

The v1 upgrader ported over and pretty much works and upgrades v0 graphs perfectly for the most part. Of course there's probably always room for improvement.

## Checklist


handsomematt:

PR #10119, opened 2026-03-04
Fix addon libraries editor not being loaded

PolEpie:
# Pull Request

---

## Summary

I use addon project for my map while developping my gamemode into the main project aside, i would love to develop and have editor tools for my mappers without needed them to map on my main gamemode project

## Motivation & Context

 so i first opened a issue #10111 but then tried with claude and found a way to getting it loaded

Fixes: #10111

## Checklist


CarsonKompon:
Other than the whitespace/formatting I think this is fine.

But it makes me wonder why we have the check for these types in the first place. What is/are the type(s) that we don't want it doing certain actions for? Why do we need to check...
```
if ( Config.Type == "game" || Config.Type == "library" || Config.Type == "addon" )
```
...if "game", "library", and "addon" ARE the 3 types? Is there another type I'm missing... or do we just not need those `if` statements at all anymore?

PolEpie:
Well idk why i listen to copilot suggestion for formatting x)
Else the type could be either:
Game, Library, Addon or Tool so the condition seem good for me 
btw i wonder what does tool ?

PR #10126, opened 2026-03-05
[Shadergraph] Tweak ParameterUI so that certain options are not shown on certain parameter nodes.

QuackCola:
## Summary

Currently There are two and a half options that show up on parameter nodes where they really don't make sense to be selectable/setable. This pr fixes that for the most part.

## Motivation & Context

To hide away settings that would either result in an error in the material editor and or wouldn't do anything on the resulting parameter.

Fixes : 
- Being able to set `UIType` to `Color` on parameter nodes that are not the Float4 ( aka color ) node. 
 Which would result in an an `Wrong Variable Type` error in the material editor like so : 
 
sbox-dev_UgJMFcP21M

- Being able to set `step` and or `UIType` on a bool parameter through the branch node.
- Being able to set `step` on the Color node.

## Implementation Details

Different ParameterUI Struct changes  : 

There is now an `IParameterUI` interface that's implemented by three different 
ParameterUI structs 
- `GenericParameterUI` Generic ParameterUI with the bare minimum amount of settings. No `UIType` option or `Step` option.
- `ColorParameterUI` ParameterUI thats used only by the Color node with no `Step` option.
- `FloatParameterUI` used by Float, Float2, and Float3 nodes. Has both a `UIType` option and a `Step` option.

I could have messed around with `ShowIfs` but personally in the case of the `ParameterUI` struct stuff its just easier for each case to have its own struct.

UIType option hiding changes :

For hiding the `UIType.Color` option for anything but the Float4 node. 
I did have to modify `EnumControlWidget.cs` and `GroupButtonControlWidget.cs` in the engine `Sandbox.Tools` project so that
you can control witch enum option could be shown through an overridden function on `EnumControlWidget` 
called `bool ShouldShowEntry( EnumDescription.Entry entry )` that's implemented in `UITypeEnumControl.cs`.

Please do let me know if there's anything at all that can be done better.

Options on a Float2 node :

Notice the `Color` `UIType` option is now gone.

sbox-dev_WCGLKw6FCC

Options on a Color node :

Notice the `Step` option is now gone.

sbox-dev_P6kxtlwKmr

Notice the `Type` and `Step` options are now gone.

sbox-dev_NSXuqoy52c

## Checklist


handsomematt:

PR #10128, opened 2026-03-05
Add ShowIndicator property to Voice Transmitter component

lucasdesca:
## Summary

Adds a `ShowIndicator` property to the **Voice Transmitter** component and a corresponding `Application.ShowMicrophoneIndicator` API, allowing games to hide the built-in microphone indicator overlay when they provide their own voice chat UI.

## Motivation & Context

Games that implement custom voice chat UI (e.g. proximity voice with custom HUD indicators, overhead speaker icons) end up with two overlapping mic indicators — their own and the engine's built-in `MicOverlay`. There is currently no way to disable the default overlay from game code.

## Implementation Details

- **`Application.ShowMicrophoneIndicator`** — new `public static bool` property (defaults to `true`). Games can also set this directly via code.
- **`Voice.ShowIndicator`** — new `[Property]` on the Voice Transmitter component, exposed as a checkbox in the editor. Sets `Application.ShowMicrophoneIndicator` on enable, resets to `true` on disable.
- **`MicOverlay.razor`** — checks `Application.ShowMicrophoneIndicator` and returns `"inactive"` when `false`, effectively hiding the overlay. Added to `BuildHash` for proper reactivity.

image

## Checklist

### DIFF
--- a/engine/Sandbox.Engine/Application.cs
+++ b/engine/Sandbox.Engine/Application.cs
@@ -162,6 +162,12 @@ public static class Application
 #endif
 
 	/// <summary>
+	/// If set to false, the built-in microphone indicator overlay will be hidden.
+	/// Games can set this to false when they provide their own voice chat UI.
+	/// </summary>
+	public static bool ShowMicrophoneIndicator { get; set; } = true;
+
+	/// <summary>
 	/// Returns true if the microphone is currently listening
 	/// </summary>
 	public static bool IsMicrophoneListening => VoiceManager.IsListening;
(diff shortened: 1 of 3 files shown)


lolleko:
Not going to happen due to privacy concerns.
We don't want you extracting peoples microphone stream without them knowing.

DeeMoney7:
> Not going to happen due to privacy concerns. We don't want you extracting peoples microphone stream without them knowing.

Would it make sense to expose the microphone indicator as a **client-side setting (default ON)**?

Totally understand not allowing games to disable it for privacy reasons, but a player-controlled toggle could help avoid conflicts with custom HUD/voice UI while still keeping the privacy guarantee.

PR #10151, opened 2026-03-08
Added `SyncFlags.Predicted`

Remscar:
## Summary

Right now only owners can change `[Sync]` properties and there are cases where you might want to fire off an RPC to the owner, which would then cause the property to change. You can't preemptively set that value, so you're stuck waiting for the owner to respond.

Solution: Add a SyncFlags.Predicted flag so non-owners can temporarily set Sync properties. Values are only local; when authoritative data arrives from the owner, it overwrites the predicted value.

## Motivation & Context

Imagine there's a button with a number displayed above it. The button is owned by the host. I press the button as a client and I have to wait until the host gets my RPC, updates the value on their end, and then the sync value for the number gets back to me. If I have a 100ms ping to the host, that's at least a 200ms delay before i see the number go up.

With a predicted value I can increment the number on the client side and show it go up immediately.

## Implementation Details

It's pretty straightforward, if a property has `SyncFlags.Predicted` then the client can change it. It will get overridden when we receive a change from the owner.

`Interpolated` and `Predicated` can technically be used at the same time. Not sure why you would want to do this, but it's possible. However the client-set value will be interpolated as well, I went back and forth on whether it should ignore interpolation.

## Checklist


badandbest:
Holy crap prediction

Remscar:
I don't know why the variant tests aren't passing. This code has nothing to do with it.

badandbest:
I had the same issue. Not your faullt. 👍

badandbest:
What I'm saying is: make an issue about it when you can.

handsomematt:

Remscar:
@kurozael Should I bother re-opening this?

PR #10208, opened 2026-03-18
fix: box-shadow disappears when using filter: brightness

zamadye:
Fixes #10207


matekdev:
This user is a bot btw

PR #10287, opened 2026-03-27
Obsolete [MakeDirty]

nixxquality:
## Summary

This PR obsoletes the MakeDirty attribute. The rest of the Dirty code seems to be pretty useful, batching successive property sets into a single OnDirty call (hopefully), so I left that intact.

## Motivation & Context

It seems to be the consensus at Facepunch that Change and MakeDirty was a mistake, and in my opinion, if something is discouraged it should actually be marked Obsolete.

## Implementation Details

The actual code of the MakeDirty attribute is 
```cs
		p.Setter( p.Value );
		OnPropertyDirty();
```
So that is what I put in every setter as a replacement.

I did not make new unit tests for this since it was a lot of components, but I jumped into the editor and changed some stuff and things seem to be fine.

## Notes

Sandbox.Renderer.RenderOptions had a MakeDirty applied to it but it doesn't have a setter, so it was never called.

The name `Component.OnPropertyDirty` is not especially great. Perhaps it should be called `Component.MarkDirty` now?

## Checklist

### DIFF
--- a/engine/Sandbox.Engine/Scene/Components/Collider/Rigidbody.cs
+++ b/engine/Sandbox.Engine/Scene/Components/Collider/Rigidbody.cs
@@ -125,11 +125,29 @@ sealed public partial class Rigidbody : Component, Component.ExecuteInEditor, IG
 	[Property, ReadOnly, Group( "Mass" ), JsonIgnore]
 	public float Mass => _body.IsValid() ? _body.Mass : default;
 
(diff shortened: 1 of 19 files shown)


aylaylay:
This defeats the point of obsoleting it, the main point of removing it is so when a property changes, it doesn't have to update 50 other things

PR #10312, opened 2026-03-30
Add glTF/GLB model import support

oldschoola:
# Pull Request

---

## Summary

Adds glTF 2.0 (.gltf/.glb) as a mesh source format using SharpGLTF.Core, following the same workflow as FBX — files appear in the asset browser and can be converted to .vmdl via right-click "Create model...".

## Motivation & Context

glTF 2.0 is the industry-standard 3D interchange format supported by Blender, Sketchfab, Adobe, and virtually every 3D tool and asset marketplace. s&box currently has no way to import glTF files — users must convert to FBX externally first.

Fixes: #8350

## Implementation Details

- **SharpGLTF.Core** (MIT, pure C#) added as a NuGet PackageReference in Sandbox.Engine for glTF parsing
- **GltfImporter** (`engine/Sandbox.Tools/Utility/GltfImporter.cs`) handles the full import pipeline:
  - Walks the glTF scene node tree, applying world transforms to each mesh
  - Merges all primitives into a single CModelMesh with face groups per material
  - Handles TRIANGLES, TRIANGLE_STRIP, and TRIANGLE_FAN primitive types
  - Converts coordinates (glTF Y-up right-handed → s&box Z-up) and units (meters → inches)
  - Extracts embedded textures to a `materials/<model_name>/` subfolder
  - Generates .vmat files with PBR properties — supports both texture references and solid color values (`baseColorFactor`, `metallicFactor`, `roughnessFactor`)
  - Registers and compiles generated .vmat files via AssetSystem
  - Material features: translucent (BLEND), alpha test (MASK), double-sided, metalness textures
  - Feeds into the existing native CModelDoc pipeline via `CreateModelFromMesh()`
- **Asset type registration** in `assettypes.txt` for both .gltf and .glb
- **Context menu** — "Create model..." works on glTF files just like FBX/OBJ/DMX
- **Asset preview** — `PreviewGltf.cs` provides thumbnails via runtime Model.Builder
- **CreateModelFromMeshFile** intercepts .gltf/.glb to route through the C# importer
- .gltf/.glb excluded from package publishing (same as .fbx)
- SharpGLTF added to `dependency_index.json` for license compliance
- All material paths use engine-relative asset paths (via AssetSystem.RegisterFile)

## Checklist


abilliontokens:
## Skeleton/Animation Support

Couldn't get skeleton or animation data into the compiled .vmdl no matter what I tried. The current PR handles static meshes with full PBR materials which covers most use cases.

The core issue is that the native CModelDoc API doesn't expose any way to add bones or animations from managed code — it only reads them from FBX/DMX source files.

Some ideas if anyone wants to pick this up:

1. **DMX Writer** — Write the glTF skeleton and animations as DMX files, then reference them from the .vmdl. The native compiler already reads DMX. ValveResourceFormat uses Datamodel.NET for this.

2. **Native API** — Expose skeleton/animation node creation in the CModelDoc managed API, similar to how CModelMesh/RenderPrimitiveMesh works now. Cleanest solution but needs engine changes.

3. **FBX Conversion** — s&box already ships libfbxsdk.dll. A managed wrapper could convert glTF to FBX, then InitFromMesh() handles everything natively.

PR #10323, opened 2026-03-31
Supress CS8974 warning & minor UI fixes to get rid of errors in Base Library in Rider

FylmTM:
# Pull Request

---

## Summary

- Add `CS8974` to list of supressed warnings (false-negative for Razor integration in Rider)
- Minor bugfix for unreacheable path validation in NavigationHost
- Use `Panel` instead of `div` when `@ref` is required in Base Library

## Motivation & Context

When opening game project in Rider it would complain about several Razor files in Base Library, marking the whole tree as having errors:
image

Specifically these 2 errors:
* `Converting method group 'OnClick' to non-delegate type 'object'. Did you intend to invoke the method?`
* `Cannot convert source type 'ElementReference' to target type 'Sandbox.UI.Panel'`

After these fixes they are gone:
image

## Implementation Details

I have chosen to add `CS8974` to the list of supressed warnings for all generated projects, making a decision that this warning is not common/popular enough, while having great dev experience is more important.

Also I have opened [issue for Rider directly](https://youtrack.jetbrains.com/issue/RIDER-137383/Razor-files-opened-in-editor-Report-error-ref-field-event-handlers-while-project-successfully-builds), hopefully someday they will introduce proper support for specifying different base class for common elements.

I have also thought about adding shims for most common html element to shadow built-in html elements, like so:
```csharp
namespace Sandbox.UI;
class partial div : Panel;
// and so on
```

It works, but decided to just fix couple places where in base library that were still using `@ref` on `div`.
Each user then can decide to create such shim in their own projects if needed.

## Checklist


handsomematt:

PR #10324, opened 2026-03-31
Move texture editor into inspector

boxrocket6803:
## Summary
Texture editor is just asset browser inspector and preview, all those things are already present in the main editor. This moves the properties from the texture editor window into the normal inspector and completely removes the texture editor tool.

## Motivation & Context
Standardizes things a bit, saves having to open a whole separate window.

## Implementation Details
There's a lot of stuff in AssetInspector set up specifically for GameResource, I've extended it slightly to allow inspectors to hijack the unsaved changes popup if they provide their own OnSave and OnReset actions.

## Screenshots
image

## Checklist


game/addons/tools/Code/Inspectors/GameResourceEditor.cs
```diff
@@ -1,6 +1,6 @@
 ﻿namespace Editor.Inspectors;
 
-[EditorForAssetType( "__fallback" )]
+[EditorForAssetType( "__fallback" ), EditorForAssetType( "vtex" )]
```
lolleko, line 3: rest of the changes look okay on first glance, but this seems like a terrible hack

boxrocket6803: best as I can tell that's the cleanest way to do it, __fallback is only for game resources, vtex isn't a game resource but should still use that editor (which is just the asset inspector in a window). There isn't really any clean way to differentiate between a non game resource asset that should use the inspector (aka just vtex) and an asset that should open in an external tool/ not open at all

PR #10325, opened 2026-03-31
Create Bootstrap.sh

Redageddon:
Matches the windows bootstrap, but for linux

# Pull Request

---

## Summary

Adds Bootstrap.sh for Linux builds.

## Motivation & Context

The batch script only works for Windows.

## Checklist


sakirma:
@Redageddon test: could you describe to me how I can bake omelette  

@Redageddon I have tried this script, however it does not work out of the box. So maybe we should add more to the Bootstrap.sh to make the project work for linux users? 

I am running the bash script with debug mode and basically the command that is being run and the error that I receive is:
```sh
+ dotnet run --project ./engine/Tools/SboxBuild/SboxBuild.csproj -- build-shaders
Error: Shader compiler executable not found at /home/hus/github.com/Facepunch/sbox-public/game/bin/managed/shadercompiler.exe
```
So I copied all binaries from the s&box downloaded from Steam and pasted it to this project bin game/bin folder. However the binary from the Steam is called ShaderCompiler.exe so I had to rename it to shadercompiler.exe

Now the next problem is making this executable work. So what would be the best idea to have shadercompiler.exe work.
Let me know what you think, maybe I am missing something and if there is a good idea then I might be able to help  us out too

Redageddon:
@sakirma sure, to make an omelette you take eggs and shove them up y-
I am aware that you can just run that script and have everything work. But the script itself is correct, and is what would be needed for a linux build. They have a lot of windows shit hardcoded into it right now (like hardcoded .exe or .dll extension names for strings and such), which would need to be changed in a different pr. But that is a different problem, not for this pr.

PR #10341, opened 2026-04-02
Add a compile definition when Disable Whitelist is on

RumBugen:
# Pull Request

---

## Summary

This PR adds a NO_WHITELIST compile definition whenever whitelist restrictions are disabled, so game code can conditionally compile behavior that is only valid in that mode. It also ensures generatet standalone solution projects receive the same symbol, keeping IDEs build-time defines aligned with compiler behavior.

## Motivation & Context

#3783 requests a compile-time symbol that can be used for #if guards when the option "Disable Whitelist" is enabled. I think that is a relatively easy task, which is also in the Tracker, to help the Facepunch team. It will also help my experiences with the engine. I didn't include tests because I couldn't find any that tested the generation of project solutions. I also couldn't find the right place to test the compilation, but I think these changes are too small to require a unit test. If I have overlooked something or a test is needed, please let me know.

Fixes: #3783

## Implementation Details

CompilerConfiguration now derives NO_WHITELIST directly from the Whitelist setting by adding the symbol when whitelist is disabled and removing it otherwise. Project.Solution also appends NO_WHITELIST for standalone-only generated solutions so solution generation matches the compilers effective define set.

-/-

## Checklist


handsomematt:

PR #10366, opened 2026-04-04
Fix glass shader tint not blending through multiple layers (#10364)

Infiland:
## Summary

Quality glass (refractive mode) writes fully opaque with no blend state, so the front glass pane completely overwrites the back glass pixels. This means only the front glass tint is ever visible when looking through multiple layers.

This PR enables premultiplied alpha blending for quality glass and computes a physically-based alpha from Fresnel reflectance and albedo luminance. Back glass layers now partially show through the front glass, blending their tints together. Clear glass (white albedo) is unaffected (alpha ≈ 1.0). Reflections remain at full strength due to premultiplied alpha.

## Motivation & Context

Fixes: #10364 

## Implementation Details

- **Premultiplied alpha** (`SrcBlend=ONE, DstBlend=INV_SRC_ALPHA`) instead of standard alpha blend — this keeps specular reflections additive at full strength while only the refraction (emission) is pre-scaled by alpha.
- **Alpha formula**: `saturate(avgFresnel + (1 - avgFresnel) * sqrt(albedoLuminance))`. At grazing angles Fresnel pushes alpha to 1.0 (glass is reflective/opaque). At normal incidence, darker tints get lower alpha so more of the back glass shows through. Clear glass gets alpha ≈ 1.0 (no visual change).
- The original albedo is saved before the quality glass path modifies it, so the luminance calculation uses the unmodified tint color.
- Cheap glass path is unchanged.
- Single-pane tinted glass will appear slightly less saturated since emission is scaled by alpha. This is a trade-off to enable multi-layer blending.

-/-

## Checklist

### DIFF
--- a/game/addons/base/Assets/shaders/glass.shader
+++ b/game/addons/base/Assets/shaders/glass.shader
@@ -78,10 +78,18 @@ PS
     // Attributes ------------------------------------------------------------------------------------------
 
     // Transparency
+    RenderState(BlendEnable, true);
     #if (S_CHEAP_GLASS)
-        RenderState(BlendEnable, true);
         RenderState(SrcBlend, SRC_ALPHA);
         RenderState(DstBlend, INV_SRC_ALPHA);
+    #else
+        // Premultiplied alpha: refraction is pre-scaled by alpha, reflections stay additive.
+        // This allows multiple glass layers to composite their tints correctly.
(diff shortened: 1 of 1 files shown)


sampavlovic:
Pretty clever, makes sense, I need to test it proper

PR #10398, opened 2026-04-08
Fix UI transform:scale clipping children at original bounds (#10162)

Infiland:
## Summary

When a UI panel uses transform: scale(...), child elements that extend beyond the original (unscaled) bounds get incorrectly clipped, even though they're visually within the scaled container.

## Motivation & Context

Panels with overflow: hidden and a scaled child would clip at the pre-transform bounds instead of the scaled bounds. This made transform: scale unusable for many UI patterns where children need to fill the scaled area.

Fixes: #10162

## Implementation Details

The GPU scissor pipeline was baking the clip-defining ancestor's GlobalMatrix into ScissorTransformMat, but the shader needs the drawing panel's own GlobalMatrix to correctly undo the transform and map screen-space pixels back to layout space for the clip test.

When a panel inherited a scissor from an ancestor with a different transform (e.g., a container with overflow: hidden and no transform, with a child at scale: 1.2), the matrix mismatch caused content near edges to be clipped at the original unscaled bounds.

The fix overrides ScissorGPU.Matrix with the drawing panel's GlobalMatrix at both scissor bake points in PanelRenderer.cs and PanelRenderer.Draw.cs, so the shader's ScissorTransformMat always pairs correctly with the panel's TransformMat.

</>

## Checklist


handsomematt:

PR #10433, opened 2026-04-09
Copy default action set to standalone VR projects.

Loafiat:
# Pull Request

---

## Summary
Fixes issue where the default action set wouldn't be copied over to exported standalone projects.

## Motivation & Context

Currently when exporting a standalone VR project in S&box the project fails to open in VR, giving the error: "Failed to attach ActionSet to Session". This PR fixes this by including the actionsets in the export if the project supports VR.

## Implementation Details
I was debating converting CoreWhitelist to an IEnumerator or adding an empty string which would be set to the right path if your project is in VR mode, but ended up just using Linq to append the path to the array when GetCoreFiles is executed.

## Checklist


Loafiat:
Not sure why formatting failed, I checked through and it seems to be all good.

PR #10446, opened 2026-04-11
Implemented repeatable action feature from Hammer into scene mapper.

Ajay-Ghale-sudo:
# Pull Request

---

## Summary

Implements repeat last action feature that was lost in translation from Hammer tool to scene view mapper. Captures a single change to a face/edge/vertex/object and allows it to be repeated by pressing a hotkey.

## Motivation & Context
I missed the repeat last action tool in the migration from the Hammer tool to the scene mapper version, so I reimplemented it. First time open source contributor here, if there are major issues with my implementation, let me know.

Fixes: #10241

## Implementation Details
Pending/Commit pattern on writes because mesh tool Record happens every frame during a drag, creating cumulative deltas
IsReplaying guard to prevent overwriting stored actions with themselves
Still not sure about the rotation implementation. It would be as simple as the rest of it if I didn't have to factor in pivots. I initially was just using a TransformDelta struct that tracked the transform from the first/most recent interaction with the gizmo to the last, but that ended up causing problems with the position of the object as it was being tracked too, and it would send them off into the distance.
Two paths: since object-mode tools don't share a common transform interface, I ended up splitting this in two or I would have had to do some major rewrites (as far as I could tell), which seemed beyond the scope of this PR.
Undo scope: RegisterUndoListener hooks OnUndo to Clear so that undoing the original action also invalidates the stored repeat data, otherwise it would be possible to undo a move and replay a stale delta on an object/face/edge/vertex.

Problems:
Currently for object mode, rotating around a changed pivot is not implemented. I tried a few methods, doing a save/restore around duplication, but OnSelectionChanged still ran and called ClearPivot mid-drag. Skipping OnSelectionChanged and manually doing it causes the OnMeshSelectionChanged to reset the pivot one frame later. If anyone has any ideas for this one, I'm all ears.

Object scale repeat action:
https://github.com/user-attachments/assets/80d493b6-9fa0-4c5a-a5ee-9f7c55e4394a

Object move repeat action:
https://github.com/user-attachments/assets/9e27a893-46d8-458e-b576-4af511a44a04

Object rotate repeat action:
https://github.com/user-attachments/assets/7979ab3b-2c85-45d5-b7ea-65cdb3acb123

Faces scale repeat action:
https://github.com/user-attachments/assets/56df8ff3-c73e-47f4-a925-d696a4f35623

Face rotate repeat action:
https://github.com/user-attachments/assets/2348a646-eed4-41fd-8ef7-18025f5ab4db

Face move repeat action:
https://github.com/user-attachments/assets/0f31c656-7ce7-459b-8be7-df5b8225c3e8

## Checklist


Ajay-Ghale-sudo:
I thought about trying to implement Trenchbroom's stack system to track a history of actions and let the user repeat multiple at a time, but that's beyond the scope of this PR I think.

PR #10523, opened 2026-04-17
Fix Linux paths unable to be found if uppercase in base file systems

joshuascript:
**Disclaimer!
Does not affect Windows behavior, but LocalFileSystem needed to have root Path manually checked against OS's. I can change this, but not sure what is recommended for that file.**

# Pull Request

---

## Summary

On Linux filesystems, Zio cannot find uppercase paths because of hardcoded lowercase-only variants. The game would only post to the LocalFileSystem call where it could find the root path or any uppercase folders in the addons/ directory. This fixes that up until a failure to initialize MenuSystem, which is a separate issue.

## Motivation & Context

Why is this change needed?
CreateAndMount calls in MenuDll are usually hardcoded to lowercase paths, and any FindDirectory lookups will regress to being unable to find uppercase directories because Linux is case sensitive. To allow for folders to be found regardless of case sensitivity, a new method was added to BaseFileSystem to segment walk the path up to the First Or Default variant using BaseFileSystem's internal method FindDirectory.

Fixes: 

## Implementation Details

There were three implementations I had tested before. I originally would tree the entire SubSystem directory and then add it to a Dictionary so look ups could be O(1) time, but this would be very hard to properly manage and would require refactoring other methods to properly update the dictionary if a file or folder was written or deleted.

The second option was just having Linux convert all directories and files to lowercase-only on runtime, but this caused a regression in several places in the engine.

The third solution, **this solution**, takes a partial path and then traverses the local root to the child folder, and compares the lowercase variant to the FirstOrDefault uppercase variant folder in the segment, and then rebuilds the folder path to get the correct one. It takes about 2-5 cycles to do this, but is the most real-time solution and does not require caching any directories to find them on launch. It also uses internal file methods in BaseFileSystem.

Attached is a screenshot of a log, where folders are returned to their uppercase variants. It works for folders only, which was the critical issue with mounted files. There are a few uppercase files handled by MenuSystem.
Screenshot from 2026-04-16 20-23-30
This fixes pathing up until MenuSystem.
Screenshot from 2026-04-16 20-25-43
This will not load textures or resource properly. This is handled by EngineFileSystem on Native Engine calls. This is a separate issue. Furthers the load to MenuSystem, where a separate filesystem issue is occurring.
image
2026/04/16 20:33:42.5506	[Generic] Error opening stylesheet: /styles/rootpanel.scss (File not found)	System.IO.FileNotFoundException: File not found
File name: '/styles/rootpanel.scss'
   at Sandbox.UI.StyleSheet.UpdateFromFile(String name, Boolean failSilently, GlobalContext ctx) in /home/joshua/Documents/GitHub/notvibecoded/sbox-public/engine/Sandbox.Engine/Systems/UI/Styles/StyleSheet.cs:line 97
2026/04/16 20:33:43.0006	[TypeLibrary] Unknown Type MenuSystem

https://pastebin.com/7Fz2H3qu <- rg "ResolveExt4" log for folder finds.

## Checklist

- [ ] Code follows existing style and conventions

### DIFF
--- a/engine/Sandbox.Filesystem/BaseFileSystem.cs
+++ b/engine/Sandbox.Filesystem/BaseFileSystem.cs
@@ -586,7 +586,7 @@ public class BaseFileSystem
 	/// </summary>
 	internal BaseFileSystem CreateAndMount( string path )
 	{
-		var sub = new LocalFileSystem( path );
+		var sub = new LocalFileSystem( FixPath ( path ) );
 		Mount( sub );
 		return sub;
 	}
@@ -594,8 +594,10 @@ public class BaseFileSystem
 	/// <summary>
 	/// Zio wants good paths to start with '/' - so we add it here if it isn't already on
 	/// </summary>
-	internal static string FixPath( string path )
+	internal string FixPath( string path )
 	{
+		if ( OperatingSystem.IsLinux() ) path = ResolveExt4FilePath( path );
+
 		// Do not allow 0-32 ASCII stuff
 		if ( path.Any( c => char.IsControl( c ) ) ) throw new ArgumentException( "Path cannot contain control characters!", "path" );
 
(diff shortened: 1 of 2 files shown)


joshuascript:
I have no idea how to do a proper Unit Test, but I provided logs for folder finds.

joshuascript:
It's fucked.

PR #10525, opened 2026-04-17
Common 3D Viewport navigation (editor)

FabiPhun:
## Summary

Rewrote the 3D Viewport Navigation within the S&box editor, to be closely resembling common 3D Software while keeping the "WASD" and camera "look around" functionality (flying camera in short) fully functional.

## Motivation & Context

 I have posted a feature request on the S&box GitHub discussing the poor navigation.
Having talked to other members on the S&box Discord, people seemed quite enthused with the idea of a navigation experience resembling most common 3D Software, I therefore created a JS prototype within a week as a proof of concept.

Fixes: 
- Merged the Pan-logic, from FirstPersonCamera with OrbitCamera
- Rewrote Orbit / Pan / Zoom logic to use distance aware navigation
- Removed unused / illogical settings within the Scene View settings
- Added new settings to support the new Orbit / Pan / Zoom

## Implementation Details

 
Files edited:
- SceneViewportWidget.cs
- SceneEditorExtensions.cs
- EditorPreferences.cs
- PageSceneView.cs
-----------------------------------------------------------
- Works in both Perspective and Orthographic Views
- Point Specific Navigation
A ray is shot from mousePos/Screenspace, when it hits something it uses the distance and position to change the zoom speed, the point around which the camera orbits and it gives the pan the necessary data to stick the point where the user clicked to the mouse cursor ( includes FOV calculations' ).

- SceneViewportWidget.cs
I had quite the fight with some of the specific exception logic and errors while adapting the new Inputs - I did some hacks here and there and would recommend the FP to find a new approach to using exceptions and Inputs within the viewport.

- Navigation Styles
I added the option for Software specific Navigation Styles instead of just one set style, the reasoning here is so anyone can use their favorite Style of Navigation.

Showcase Video:
https://github.com/user-attachments/assets/fe579fe6-38bc-4a66-978f-1213164f0b44

Settings Screenshot:
Screenshot_4
# Pull Request

##Checklist


handsomematt (changes requested):
There's like 800 changed files here, I really don't like the navigation style stuff we should have our own style and that's it.

FabiPhun:
> There's like 800 changed files here, I really don't like the navigation style stuff we should have our own style and that's it.

The changes are due to a merge that had to be forced caused by all PR's being closed twice on the master branch.
Request to review new PR https://github.com/Facepunch/sbox-public/pull/10589.
I am 100% on board to adjust and modify my code to fit everyones need.

PR #10542, opened 2026-04-17
Include library and transient assets in file networking

giodotblue:
## Summary

Sends library and transient assets to connecting clients

## Motivation & Context

Library assets aren't sent when using the "Join from new instance" button, but the assets work fine locally

Similarly, game transients aren't getting sent, so compiled SVGs etc aren't working on the new instance (unless the game has already been uploaded to sbox.game)

## Implementation Details

I've avoided changes to asset / engine code here, and the only change outside of GameInstance.Network.cs is a new pair of internal methods in Project (which aren't fully necessary, but nice to have)

This PR might be something that could be solved somewhere else (maybe the game FS should be mounting transients?) but I'm not sure

### DIFF
--- a/engine/Sandbox.Engine/Systems/Project/Project/Project.cs
+++ b/engine/Sandbox.Engine/Systems/Project/Project/Project.cs
@@ -211,6 +211,16 @@ public sealed partial class Project
 	/// </summary>
 	public bool HasAssetsPath() => RootDirectory is not null && System.IO.Directory.Exists( GetAssetsPath() );
 
+	/// <summary>
+	/// Absolute path to the .sbox/transient folder of the project, or <see langword="null"/> if not set.
+	/// </summary>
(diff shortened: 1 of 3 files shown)


handsomematt:
Isn't all this stuff all in FileSystem.Mounted or something?

gio3k:
Yes, but the original code uses GameInstance.GameFileSystem directly (I assume it was so only game related files are sent, but that's not really working out for us)

I could rewrite this PR to use FileSystem.Mounted if you'd like? I assumed there would be a reason it's not doing so already

gio3k:
Oh, it's because FileSystem.Mounted has 15k files, we'd be sending thousands of assets from base / core - BuildNetworkedFiles is for the initial file package, so it just sends everything under the include filter

That's off the table then, this PR could stay the same or it could be rewritten to add a GameInstance.Mounted with just game fs + library fs + transients? What do you think @handsomematt

lolleko:
You still want to go ahead with this @giodotblue ?

giodotblue:
Sure, that's up to you guys

PR #10473, opened 2026-04-14
Add middle-click reset functionality to FloatControlWidget

Tripperful:
## Summary

Middle click any FloatControlWidget in the editor to reset it to default value (0).

## Motivation & Context

Very handy, 3ds max does this

## Screenshots / Videos (if applicable)

https://github.com/user-attachments/assets/35c737bd-130e-4604-ae4c-8ccd93b4979d

## Checklist

- [x] Code follows existing style and conventions
- [x] No unnecessary formatting or unrelated changes
- [x] Public APIs are documented (if applicable)
- [x] Unit tests added where applicable and all passing
- [x] I’m okay with this PR being rejected or requested to change 🙂

PR #10559, opened 2026-04-18
Add middle-click reset functionality to FloatControlWidget

Tripperful:
## Summary

Re-submit of #10473

Middle click any FloatControlWidget in the editor to reset it to default value (0).

## Motivation & Context

Very handy, 3ds max does this

https://github.com/user-attachments/assets/35c737bd-130e-4604-ae4c-8ccd93b4979d

## Checklist

### DIFF
--- a/engine/Sandbox.Tools/ControlWidget/FloatControlWidget.cs
+++ b/engine/Sandbox.Tools/ControlWidget/FloatControlWidget.cs
@@ -321,6 +321,14 @@ public class FloatControlWidget : StringControlWidget
 			dragging = false;
 			PropertyFinishEdit();
 		}
+
+		if ( e.MiddleMouseButton && !ReadOnly )
+		{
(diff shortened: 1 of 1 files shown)


joshuascript:
This is an excellent feature. Does the editor key track of key controls anywhere? This looks a little hard coded in, but if this was a control feature that could be mapped to a key I could completely get behind that.

Tripperful:
@joshuascript this is implemented in line with the rest of the file, i. e. up-down arrows, left, right click are all hardcoded there

handsomematt (commented):
Would you not want this implemented on all control widgets?

Tripperful:
@handsomematt as there are currently 84 different control widgets, I am not sure it makes sense for all of them to have this reset functionality. I am also not sure all of the default values that those would get reset to currently make sense.

You guys probably have a better idea, if you think it makes sense to do this in the base ControlWidget, I am completely fine with that

PR #10568, opened 2026-04-18
Linux fix Zio paths not resolving folder paths due to case sensitivity mismatch.

joshuascript:
# Pull Request

---

## Prerequisites
There is a native bug where a double-free occurs. A shim is needed to run the base game.
Make a launch.sh script -> https://pastebin.com/raw/TJSsuDxX
Make a no_double_free.c file -> https://pastebin.com/raw/9pcnqmdy
Put both in /game folder and run ./launch.sh (try a few times if you cannot see FP logo) 
Also puts libskiasharp and libharfbuzz in managed folder, which is correct order needed.

If you see "Couldn't Create MenuSystem!" that means code is working. To get to missing texture screen, rename:
/addons/base/code -> /addons/base/Code
/addons/tools/assets -> /addons/tools/Assets
This is a separate file path issue in Project.cs from Sandbox.Engine.

## Summary

This fixes filepath issues that occur in `Sandbox.FileSystem`, where roughly 200 incoming file paths mounted
through Zio are lowercase-only. This issue happens in *CreateSubSystem* when the game is
launched (i.e. *CreateSubSystem()* received *addons/citizen/assets* which cannot resolve
*addons/citizen/Assets*).

## Motivation & Context

This builds the engine to work natively on Linux. This is an incremental change that is part of 3 other ones. The game can run natively if the file path resolutions can be met properly. This change gets the filesystem a third of the way there. The original plan was to cache directories in a SubSystem by walking the entire tree for each mount, but it would be hard to keep track of, and this issue only occurred for about 200 file paths out of 2000 in the base addons.

An alternative solution was to make the entire Linux filesystem be lowercase and compare uppercase directories to lowercase variants, but it regressed the engine in several places and is not worth it for such a small issue in Zio.

## Implementation Details

Revised with Claude

* *ResolveLinuxFilePath()* added to *FixPath()* in *BaseFileSystem.cs* — corrects incoming paths for all existing fs methods
* *FixPath()* is no longer a static method because it is only used internally in `Sandbox.FileSystem` and is required to pass non-static system variable into Linux method
* *LocalFileSystem* uses *NormalizePath()* (a variant of `NormalizeFilename`) to normalize file roots to OS-specific paths
* *FixPath()* behavior is unchanged except for a new ext4 check
* *ManageExt4.cs* added as a dedicated class for Linux-only fixes going forward

When a path doesn't exist as-is, *ResolveLinuxFilePath()* breaks it into segments and walks back up the root, enumerating child directories and matching case-insensitively until the correct path is reconstructed. Most paths resolve without this — only ~200 of ~2,000 calls need the walk.

*ManageExt4.cs* exists to maintain a clear separation between Linux and Windows concerns. The ultimate goal is to <u>never regress anything Windows-related</u>, keeping changes to *BaseFileSystem* minimal (<15 lines). It also gives a single place to handle future divergences — for example, if *CreateDirectory()* is called with */code* in one place and */Code* in another, Linux ends up with two folders while Windows has one. That kind of fix can live here and be forgotten about, without requiring Facepunch developers to audit every path choice.

**Before.**
Screenshot from 2026-04-18 02-23-01

**After.**
Screenshot from 2026-04-09 04-23-12

## Tests
This is a log of the filepaths being parsed correctly.
https://pastebin.com/raw/TaFXAgVU

## Potential Issues
I am highly confident the ERROR_FILEOPENs are not related to Zio, but rather the way that resources are sent to NativeEngine calls elsewhere in GameInstanceDll or MenuDll. I would recommend confirming that this is a standalone issue before accepting this solution. 

## Checklist

### DIFF
--- a/engine/Sandbox.Filesystem/BaseFileSystem.cs
+++ b/engine/Sandbox.Filesystem/BaseFileSystem.cs
@@ -596,74 +596,31 @@ public class BaseFileSystem
 	/// </summary>
 	internal string FixPath( string path )
 	{
-		if ( OperatingSystem.IsLinux() ) path = ResolveExt4FilePath( path );
-
 		// Do not allow 0-32 ASCII stuff
 		if ( path.Any( c => char.IsControl( c ) ) ) throw new ArgumentException( "Path cannot contain control characters!", "path" );
 
 		if ( path.Length < 1 )
 			return "/";
 
-		if ( path[0] == '/' ) return path;
-		return string.Concat( "/", path );
+		if ( path[0] != '/' ) path = string.Concat('/', path);
+
+		if ( OperatingSystem.IsLinux() )
+			path = ManageExt4.ResolveLinuxFilePath( system, path );
+
+		return path;
 	}
 
 	/// <summary>
-	/// Sometimes Linux filesystems cannot be resolved because the engine
-	/// attempts to compare a lowercase path to an uppercase one. This rebuilds
-	/// the folder and file paths in segments to allow the engine to build the uppercase
-	/// variant. This only applies to Zio mounts and will not fix managed calls to the native
-	/// engine.
+	/// Passes through case handled paths for Zio system root paths
 	/// </summary>
-	internal string ResolveExt4FilePath( string path )
+	/// <param name="path"></param>
+	/// <returns></returns>
+	internal string NormalizePath( string path )
 	{
-		if ( system is null ) return path;
+		path = path.Replace( "\\", "/" );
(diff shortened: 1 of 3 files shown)


joshuascript:
Happy to change anything or create a specific Unit Test if needed. Logs provided were taken from ResolveLinuxFilePath itself during testing.

joshuascript:
Got feedback. Changes soon.

handsomematt:
ResolveLinuxFilePath looks like a horror show

joshuascript:
Back to the drawing board.

PR #10571, opened 2026-04-18
static Game.Overlay.IsAnyOpen and Game.Overlay.IsPauseOpen

boxrocket6803:
## Summary
2 properties in Game.Overlay that could be static, are referred to as static, and make much more sense being static.
```cs
Game.Overlay.IsOverlayOpen
Game.Overlay.IsPauseOpen
```

previously
```cs
var overlay = new Game.Overlay();
overlay.IsOpen
overlay.IsPauseMenuOpen
```

## Motivation & Context
Didn't feel like keeping track of an instance of Game.Overlay

## Implementation Details
Seemed I should obsolete rather than just make the existing properties static to avoid breaking things. New names aren't quite as good though.

## Checklist


handsomematt:
Probably makes sense, will dwell on if it's the right place

PR #10620, opened 2026-04-21
Add sounds to NS2 mount

nixxquality:
## Summary

This PR adds the missing sound files to the NS2 mount.

## Motivation & Context

I would like to use the NS2 sounds together with other NS2 assets.

## Implementation Details

The sound files in NS2 are grouped together in FMOD Sound Banks.
I based my implementation on https://github.com/HearthSim/python-fsb5

NS2 only uses MPEG and PCM16 formats inside the banks.

When NS2 is mounted, it will read the header of every bank and add a loader to the context for each sample inside.
The loader will then read inside the bank file at the correct offsets.

What makes this a WIP is that I have no idea how to create sounds from just samples and settings. I'm pretty sure that's internal-only for now. Therefore, I have two approaches for the actual 'loading' part:

1. For PCM16, I create a fake .wav file in memory and load it.
2. For MPEG files I write them to disk as a proof of concept.

Another note: NS2 makes liberal use of the LOOP metadata in the sample chunks. That information is just being skipped for now.

https://github.com/user-attachments/assets/55aa581b-b25a-4cb1-bc37-550bdb6fd206

## Checklist

## Note

Remake of #10051

### DIFF
--- a/engine/Mounting/Sandbox.Mounting.NS2/GameMount.cs
+++ b/engine/Mounting/Sandbox.Mounting.NS2/GameMount.cs
@@ -22,7 +22,8 @@ public class GameMount : BaseGameMount
 	{
 		{ ".model", ResourceType.Model },
 		{ ".dds", ResourceType.Texture },
-		{ ".material", ResourceType.Material }
+		{ ".material", ResourceType.Material },
+		{ ".fsb", ResourceType.Sound },
 	};
 
 	protected override Task Mount( MountContext context )
@@ -49,6 +50,10 @@ public class GameMount : BaseGameMount
 			else if ( resourceType == ResourceType.Material )
 			{
 				context.Add( resourceType, path, new MaterialLoader( fullPath ) );
+			}
+			else if ( resourceType == ResourceType.Sound )
+			{
+				SoundBankLoader.AddSoundsFromBank( context, fullPath, path );
 			}
 		}
 
--- a/engine/Mounting/Sandbox.Mounting.NS2/Resource/SoundBank.cs
+++ b/engine/Mounting/Sandbox.Mounting.NS2/Resource/SoundBank.cs
@@ -34,14 +34,6 @@ class SoundBankLoader( string bankPath, SampleInformation info ) : ResourceLoade
 		VORBIS
 	};
 
-	static int WidthForFormat( SoundFormat format ) => format switch
-	{
-		SoundFormat.PCM8 => 1,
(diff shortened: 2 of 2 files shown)


aylaylay:
Love this, nice and simple. You should be able to create sounds from PCM though. If this is internal I'll push something to make it public

aylaylay:
I do worry that this is doing too much at the mount stage, really all the heavy lifting should be done on the loader

nixxquality:
The conclusion we reached last time is that we're missing
- [ ] A function to create a sound from PCM data
- [ ] A function to create a sound from MP3 data

I do wish we could do async things in the Mount function, but MountContext being a ref struct prevents this. Would be a breaking change to fix.
I'm not sure how much 'heavy lifting' actually happens. I can time it if you'd like, but it's just reading a small amount of data from 33 files.
Also, I'm not sure where else you would add dynamic files like this. The current mounting system only lets me do this, as far as I understand.

aylaylay:
It's fine if it has to do it, we do some work in other mounts when reading from zip files etc, it's just if more work can be moved to the loader then it should

nixxquality:
Okay. When I have the functions I need to finish this PR I'll see if I can reduce the amount of work in the mounter to the bare minimum.

aylaylay:
you can now use `SoundFile.FromPCM`

```cs
	protected override object Load()
	{
		using var bf = File.OpenRead( BankPath );

		var data = new byte[Info.Length];
		bf.Seek( Info.DataOffset, SeekOrigin.Begin );
		bf.ReadExactly( data );

		if ( Info.Format != SoundFormat.PCM16 )
			return null;

		return SoundFile.FromPCM( Path, data, Info.Channels, (uint)Info.Frequency, 16, false );
	}
```

FromMP3 soon

aylaylay:
You now have `SoundFile.FromPcm` and  `SoundFile.FromMp3`

https://github.com/user-attachments/assets/cc4a1e02-8a90-4e52-957e-b6ed9f1d31f0

nixxquality:
https://github.com/user-attachments/assets/6f499260-b093-4947-b20b-37eb6d4effb8

As requested, I also moved some stuff out of the Mount function, but there's still a bit of code there.

I started the editor with Superluminal and checked the total runtime of AddSoundsFromBank: **25ms**.
NS2 is installed on one of my harddrives, but the fsb files may still be in cache from previous runs.

PR #10635, opened 2026-04-22
Register dropped files as assets

pondepth:
Fixes an issue where external files dropped into the asset browser were copied into the project but not registered with AssetSystem until editor restart.

Changes:
- Register dropped files after copy/move
- Handle dropped folders recursively
- Lazily register plain file entries before opening the asset context menu


handsomematt:
This feels so wrong. What happens when you put them in your project folder in Windows Explorer? I'd expect them to be found and registered automatically, does that not happen? Why would this be different from that?

PR #10650, opened 2026-04-23
Add custom code editor option with browse button

lucasdesca:
## Summary

Adds a browse button ("...") next to the code editor dropdown in preferences, letting you manually point to any editor executable. Useful when auto-detection fails, the editor isn't one of the bundled options, or you have a custom install.

## Motivation & Context

The current dropdown only lists Visual Studio, Rider, and VS Code, detected via the Windows registry. This doesn't cover:

- Editors not on the list (Sublime, Fleet, Zed, Vim/Neovim, Emacs, etc.)
- Portable / non-standard installs that skip registry entries
- Users on Linux, where registry-based detection doesn't apply

This PR is a simple fallback for all those cases.

## Implementation Details

- New `CustomCodeEditor` class implementing `ICodeEditor`. Path persisted in `EditorCookie` under the key `CodeEditor.CustomPath`.
- Browse button added to `CodeEditorControlWidget` using `IconButton` + `FileDialog` — same pattern as `FilePathControlWidget` in MovieMaker.
- `FileDialog` filter is platform-aware: `*.exe` on Windows, `All files (*)` elsewhere.
- `CustomCodeEditor` is intentionally excluded from auto-discovery in `CodeEditorControlWidget`; the entry only appears once the user has picked a path, and it's labeled with the selected exe name so multiple picks are distinguishable.
- `line` / `column` parameters on `OpenFile` are intentionally ignored (documented with an XML comment) — the jump-to-line argument syntax varies per editor (`path:line:col`, `-n{line}`, `+line`, ...) and can't be predicted for an arbitrary executable.

## Screenshots / Videos

https://github.com/user-attachments/assets/8b44ff0d-e338-4fe8-8e6e-acefbe7f8aea

## Checklist

- [x] I'm okay with this PR being rejected or requested to change 🙏

### DIFF
--- a/game/addons/tools/Code/CodeEditors/CodeEditor.Custom.cs
+++ b/game/addons/tools/Code/CodeEditors/CodeEditor.Custom.cs
@@ -1,0 +1,52 @@
+namespace Editor.CodeEditors;
+
+/// <summary>
+/// Opens source files with a user-picked executable. The path is persisted in
+/// an editor cookie and is configured via the browse button on the code editor
+/// control. Useful as a fallback when Visual Studio / Rider / VS Code aren't
+/// installed, or for any editor not covered by a dedicated <see cref="ICodeEditor"/>.
+/// </summary>
+[Title( "Custom" )]
+[Description( "Manually selected code editor" )]
+public class CustomCodeEditor : ICodeEditor
+{
+	public static string CustomPath
+	{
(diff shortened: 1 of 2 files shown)


wpenistone:
Similar PR down here, see latest revision:
https://github.com/Facepunch/sbox-public/pull/9995

PR #10682, opened 2026-04-26
.sh files should be LF

DrakeFruit:
## Summary

adds .sh files to the list of exceptions for auto formatting, always use LF

## Motivation & Context

without this, every user has to run dos2unix ./*.sh when they first clone the repo.

## Screenshot

image

## Checklist


handsomematt:
Treat the .git* files as read only

PR #10715, opened 2026-04-28
Fix lasso inconsistency by using mouse down instead of was down

glimpsebeyond:
# Pull Request
## Summary

While using mapping tools I needed to lasso heavily from the top down view in face mode. The lasso would fail to activate > 50% of the time. I think this is likely related a race condition around Gizmo.WasLeftMousePressed or possibly the modifier keys. But since that's harder to fix I put in a small workaround that doesn't seem to change the lasso behavior but makes it 100% reliable. I've been using it for a week or two with no issues. Ideally we'd fix the root but this works for now.

## Motivation & Context
Lasso not firing drives a man insane.

## Implementation Details
This is a bandaid. Fixing the root is preferable.

## Checklist

### DIFF
--- a/game/addons/tools/Code/Scene/Tools/EditorTool.cs
+++ b/game/addons/tools/Code/Scene/Tools/EditorTool.cs
@@ -4,16 +4,11 @@ namespace Editor;
 
 public class EditorTool : IDisposable
 {
-	[JsonIgnore]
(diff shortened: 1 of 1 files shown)


glimpsebeyond:
I didn't want to mess with all that whitespace but formatting wasn't passing. Not sure how the file wasn't already formatted correctly.

PR #10866, opened 2026-05-07
Display a reason when a Http request is blocked

nixxquality:
## Summary

This PR adds more feedback to the Http.IsAllowed function than a simply boolean can provide. This will improve debugging in the future and save developers time.

## Motivation & Context

Someone in the discord was wondering why requests to his OVH-provided domain name were blocked. It seems that they were being resolved to a local address, and thus blocked for security reasons, but this was not clear from the error message.
In order for his problem to be fixed (by added `-allowlocalhttp` to his command line arguments) he first had to figure out *what* was going wrong.
By having more information available in the initial request for help, this could have saved everyone time.

## Implementation Details

IsAllowed now returns a struct which implicitly converts to bool for backwards compatibility.
All functions which use Http.IsAllowed now print the error reason which has been added to every case.

```cs
		var isAllowed = Http.IsAllowed( request.RequestUri );
		if ( !isAllowed )
			throw new InvalidOperationException( $"Access to '{request.RequestUri}' is not allowed. {isAllowed.Reason}" );
```
I'm not too happy with this pattern, so if you have any ideas on how to write it shorter, let me know.

Sample project (just has a `request` concmd which executes an Http.RequestAsync): [reqtest.zip](https://github.com/user-attachments/files/27484631/reqtest.zip)

For some reason, matt's 127.0.0.1 domain does not resolve for me (I'm 99% sure it's a local network issue, don't worry about it), so I tested with a domain which my router resolves to itself to make sure that works properly:

Because of this possible issue, I wrote the unit test to be ambiguous and accept both errors (just check if it contains to word 'resolve')

## Checklist


nixxquality:
Pretty sure the tests failed because of an unrelated issue.

engine/Sandbox.Engine/Utility/Web/Http.cs
```diff
@@ -39,6 +39,35 @@ static Http()
 	/// </summary>
 	internal static bool IsLocalAllowed => ((Application.IsEditor || Application.IsDedicatedServer) && CommandLine.HasSwitch( "-allowlocalhttp" )) || Application.IsStandalone;
 
+	/// <summary>
+	/// Can be used like a boolean to see if a Uri can be accessed.
+	/// If not, there should be a human-readable reason to display.
+	/// </summary>
+	public struct IsAllowedResult
```
lolleko, line 46: I would be happy with more explicit throws but dont want a public struct like this.

lolleko (changes requested):

handsomematt:
Returning an error like this is a big c# anti pattern

PR #10881, opened 2026-05-08
Fix editor DPI maximize window issue

PolEpie:
# Pull Request

---

## Summary

So when maximizing editor window, most of times it get stuck like that 1/2 times
image

## Motivation & Context

So annoying to try maximize 2/3 times

Fixes: #1651

## Implementation Details

So the issue is that there was 2 recording logic
Lock condition (line 107): IsRecording() && sceneCamera.IsRecordingCamera == true
Unlock condition (line 115): !IsRecording() && _sizeLockedForRecording

So i changed to do:
// before:
else if ( !ScreenRecorder.IsRecording() && _sizeLockedForRecording )
// after:
else if ( _sizeLockedForRecording && (!ScreenRecorder.IsRecording() || sceneCamera?.IsRecordingCamera != true) )

## Checklist

- [ ] Code follows existing style and conventions
- [ ] No unnecessary formatting or unrelated changes
- [ ] I’m okay with this PR being rejected or requested to change 🙂

### DIFF
--- a/engine/Sandbox.Tools/Qt/SceneRenderingWidget.cs
+++ b/engine/Sandbox.Tools/Qt/SceneRenderingWidget.cs
@@ -34,6 +34,25 @@ public class SceneRenderingWidget : Frame
 	private Vector2 _savedMinSize;
 	private Vector2 _savedMaxSize;
 
+	/// <summary>
+	/// Releases the recording size lock on all active SceneRenderingWidgets.
+	/// Called from <see cref="Window.OnWindowStateChange"/> because Qt only delivers
+	/// WindowStateChange to the top-level window, not to child widgets.
+	/// Without this, a lock left active while the window is minimized would keep
+	/// the min/max size constraint in place on restore/maximize, causing layout and
+	/// swap-chain issues.
+	/// </summary>
+	internal static void ReleaseAllRecordingLocksForWindowStateChange()
+	{
+		foreach ( var w in All )
+		{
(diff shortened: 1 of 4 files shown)


PolEpie:
Ok so the first commit was just not enought and i just wrongly tried it,

So the real issue happen when OnWindowStateChange get called, in QT it call interally a OnResize that set to a non DPI scaled size my window and kinda fuck up everything

So what i did is that when OnWindowStateChange is firing ima cache the window value, and when OnResize launch and i have cache we override the size to not broke the window and avoid maximize and have small window

I know the code is not perfect but it's a forward path to get a better window management

PolEpie:
Else for me the issue might be deeper in QT, there is severals things i saw,

When you click twice on the icon of sbox editor on your taskbar, you'll get the weird window issue,
If you spam the icon sbox editor, like 4~5 times you hear a windows error song playing and the window doesnt open
When starting the editor never get fullscreen for me

(BTW i have DPI scaling at 225%)

i find that it could be QWidget::restoreGeometry() and how it interacts with WINDOWPLACEMENT.rcNormalPosition on Windows at non-100% DPI.

I tried dig with claude and he said that rcNormalPosition wasnt saved well with DPI
and since the window start maximized, it doesnt help and never get it corrected (this part might be wrong)

PolEpie:
After many hours at digging it's something from QT related to DPI bruh another one...

PR #10884, opened 2026-05-08
StyleSheet Parsing fix

obselate:
# Pull Request

---

## Summary

Calling `panel.StyleSheet.Parse(css)` repeatedly **does not replace** the previous sheet. Every call appends a new sheet to the collection, accumulating forever. Because two sheets can carry blocks for the same selector with identical scores, the cascade winner becomes non-deterministic.

Root cause: `StyleSheet.FromString` (and the `StyleParser.ParseSheet` it delegates to) never assigns `sheet.FileName`. `Parse`'s replace logic relies on `Remove("string")` matching the prior sheet by `FileName`, but `FileName` is null on every `FromString` sheet, so `Remove` is a no-op.

## Motivation & Context

`StyleSheetCollection.Parse(string)` is meant to **replace** the previously parsed stylesheet; it calls `Remove("string")` then `Add(...)`. The `Remove` step is a no-op on every sheet `Parse` has ever installed, so each call leaks the prior sheet onto the panel. After *N* `Parse` calls, the panel carries *N* sheets all matching the same selectors.

> [!IMPORTANT]
> This is a silent correctness bug, not just a leak. Leaked blocks tie on both `SelfScore` and `LoadOrder` (the latter is per-sheet, not global), so the cascade sort is unstable and which value wins is nondeterministic.

**Cause:** `StyleSheet.FromFile` sets `FileName`, `FromString` doesn't. `Remove(string)` filters by `FileName != null`, so `FromString` sheets are invisible to it.

**Hits anything that re-parses CSS at runtime:** tool addons, live previews, hot-reload loops. Game UIs typically `Load(path)` once and never re-parse, which is why this hasn't surfaced widely. I hit it building a Razor-preview tool that regenerates CSS on every Inspector edit.

### Adjacent open issues (not duplicates)

- #8397 
- #4779
- #3807

## Fix:

`StyleSheet.FromString` now sets `FileName` on the returned sheet, matching `FromFile`. This makes `StyleSheetCollection.Remove("string")` actually find and remove the prior sheet on subsequent `Parse(string)` calls, so `Parse` replaces instead of accumulates.

One-line change in `StyleParser.ParseSheet`:

```diff
 public static StyleSheet ParseSheet( string content, string filename = "none", IEnumerable<(string, string)> variables = null )
 {
     IncludeLoops = 0;

     StyleSheet sheet = new();
+    sheet.FileName = filename;
     sheet.AddVariables( variables );

     ParseToSheet( content, filename, sheet );

     return sheet;
 }
 ```

## Implementation Details

`FromFile` already assigns `FileName` next to its `AddVariables`/`AddWatcher` calls. Doing the same in `ParseSheet` keeps both factory paths symmetric.

Will just post reproduction of the issue:

Minimal repro
```csharp
using System.Linq;
using Sandbox.UI;

var p = new Panel();
p.StyleSheet.Parse( ".x { width: 100px; }" );
p.StyleSheet.Parse( ".x { width: 200px; }" );

var sheets = p.AllStyleSheets.ToList();
Log.Info( $"sheets={sheets.Count}" ); // expected 1, observed 2

foreach ( var s in sheets )
{
    Log.Info( $"  FileName=\"{s.FileName ?? "<null>"}\" nodes={s.Nodes.Count()}" );
    foreach ( var node in s.Nodes )
    {
        var w = node.GetRawValues().FirstOrDefault( v => v.Name == "width" ).Value;
        Log.Info( $"    {string.Join(",", node.SelectorStrings)} → width={w}" );
    }
}
```

Expected:
```csharp
sheets=1
  FileName="<null>" nodes=1
    .x → width=200px
```

Observed:
```csharp
sheets=2
  FileName="<null>" nodes=1
    .x → width=200px
  FileName="<null>" nodes=1
    .x → width=100px
```

## Checklist


Mr-VIT:
You can pick my commit https://github.com/Mr-VIT/sbox-public/commit/a2b97c512844d36f0371898c789663faf098ac45
`Remove()` after `CollectVariables()`
Add `:` to the file name to skip FS checks
Tests passed

PR #10890, opened 2026-05-09
Fix initial project load opening blank scene instead of minimal.scene

aidencurtis:
# Pull Request

  Thanks for contributing to s&box ❤️

  ———

  ## Summary

  Fixes an editor layout issue where new projects could open with an untitled scene instead of the generated scenes/minimal.scene.

  When restoring the default editor layout, the editor now prefers the configured StartupScene when it exists and falls back to scenes/minimal.scene when it does not. This also keeps Reset Layout consistent with first-time default layout restore.

  ## Motivation & Context

  New projects generate a minimal scene, but the default editor layout could still restore an untitled scene tab. That made new projects look like no scene had been created or selected.

  This also meant manually restoring the default layout could reopen the wrong scene state instead of the expected startup/default scene.

  Fixes: https://github.com/Facepunch/sbox-public/issues/418

  ## Implementation Details

  - Updated EditorMainWindow default layout restoration to resolve an initial scene path before applying the layout.
  - When restoring the default layout:
      - prefer the configured StartupScene when it is valid and exists
      - fall back to scenes/minimal.scene when the startup scene is missing
      - ignore invalid or non-.scene paths
  - Kept the behavior at the layout level by replacing the default SceneDock:untitled entry in the restored layout state, so the correct scene dock is created directly.

  N/A - editor startup/layout behavior change only.

  ## Checklist

### DIFF
--- a/engine/Sandbox.Tools/Editor/EditorMainWindow.cs
+++ b/engine/Sandbox.Tools/Editor/EditorMainWindow.cs
@@ -1,9 +1,12 @@
-﻿using NativeEngine;
+﻿using System;
+using NativeEngine;
 
 namespace Editor;
 
 public class EditorMainWindow : DockWindow
 {
+	const string MinimalScenePath = "scenes/minimal.scene";
+
 	internal static EditorMainWindow Current;
 
 	Menu FileMenu { get; init; }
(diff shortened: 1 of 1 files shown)


aidencurtis:
this seems to be working now on master? closing even though I don't know what changed lol

PR #10912, opened 2026-05-11
Init EditorAppSystem with the editor app id, stops Steam overlay from appearing in editor with shift + tab.

Matt9440:
## Summary

Init EditorAppSystem with the editor app id, stops Steam overlay from appearing in editor with shift + tab. Restores the intended editor navigation functionality.

## Motivation & Context

Steam overlay should not be appearing in editor, shift + tab to navigate backwards is preferred.

Fixes: 
https://github.com/Facepunch/sbox-public/issues/9955

## Considerations
This will also stop Steam from tracking play time on the game when using the editor, instead the playtime is recorded against the editor.

## Checklist

### DIFF
--- a/engine/Sandbox.AppSystem/EditorAppSystem.cs
+++ b/engine/Sandbox.AppSystem/EditorAppSystem.cs
@@ -5,8 +5,12 @@
 /// </summary>
 public class EditorAppSystem : AppSystem
 {
+	const ulong EditorAppId = 2129370;
+
 	public override void Init()
 	{
+		Application.AppId = EditorAppId;
+
 		LoadSteamDll();
 
 		base.Init();


Matt9440:
Going to convert this back to a draft, it's evident editor playtime being recorded against the games app id is intentional for things like random drops. Will need to try to find another way to supress the overlay, though its probably best handled natively.

handsomematt:
I don't think we really care about drops happening in editor. I'd be more concerned about networking not working here. Or the auth tokens fucking up.

Why does this stop the overlay? We had a env var set for this before, but it seems to have broken.

Matt9440:
I think it's disabled by default for Software app types

https://partner.steamgames.com/doc/features/overlay

PR #10955, opened 2026-05-15
Normalize RPC NetFlags to reliable unless explicitly unreliable

Matt9440:
Summary
Normalize RPC NetFlags so RPCs remain reliable unless NetFlags.Unreliable is explicitly specified.

Motivation & Context
`[Rpc.Broadcast(NetFlags.HostOnly)]` replaces the default flags entirely, meaning the default `NetFlags.Reliable` was secretly dropped if for example you specified that only the host could trigger this broadcast.

The now obsolete `[Broadcast]` attribute preserved reliability because it OR'd permission flags onto NetFlags.Reliable.

Implementation Details
- RpcAttribute.Flags are normalized and are always Reliable unless the Unreliable flag is set.
- Unit tests confirming default reliability and explicit unreliability

Checklist
- [X] I’m okay with this PR being rejected or requested to change 🙂ted to change 🙂

### DIFF
--- a/engine/Sandbox.Engine/Scene/Networking/Rpc.Attributes.cs
+++ b/engine/Sandbox.Engine/Scene/Networking/Rpc.Attributes.cs
@@ -9,12 +9,21 @@
 public abstract class RpcAttribute : Attribute
 {
 	internal RpcMode Mode { get; set; } = RpcMode.Broadcast;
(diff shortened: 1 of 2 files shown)


MrBrax:
There was always a discussion about this, some people insisted it was reliable by default but I always add the flag because host only would override it

PR #10995, opened 2026-05-19
 Set Property & Get Property Blocks In Doo Edtior

2czu:
# Pull Request

---

## Summary

We weren't able to get or set component properties through the Doo Editor.

## Motivation & Context

Our Attack system works with Actions, and because the Action Graph will most definitely be deprecated in a few months, we want to switch to something that WILL be updated.
This is only in our case, but setting and getting properties from a component is, in my opinion, a key feature to any functional (as in made to invoke functions) editor.

Fixes: 

## Implementation Details

To be fair, there aren't any non-obvious decisions, I based my changes (adding 2 blocks, serialization, component property fetching) on what you've already done in the Doo Editor for other functionalities

image

## Checklist


handsomematt (changes requested):
I'd be happy to accept but your pull request has 100s of unrelated files

PR #11026, opened 2026-05-22
Fix null reference exceptions across engine and game code

PatrickJnr:
# Pull Request

---

## Summary

Fixes 7 potential Null Reference Exceptions across engine and game code. These mainly come from unsafe `.First()` / `.FirstOrDefault()` usage where results are assumed to exist but can be empty in real runtime or editor conditions.

In practice, these can show up during normal usage too (asset picker, terrain updates, chair exits), not just rare edge cases.

---

## Motivation & Context

These were found via a static scan looking for unsafe collection access patterns.

Some are edge cases, but several are realistically reachable during normal gameplay/editor flows when objects are missing, destroyed, or not yet initialised.

---

## Implementation Details

This is a defensive cleanup pass rather than a behavioural change.

Main approach:

* Replace unsafe `.First()` calls with `.FirstOrDefault()`
* Add null/validity checks where chained access could fail
* Split dereferences into locals where needed to avoid hidden NREs
* Keep existing logic intact unless it was already unsafe by design

### Changes

| File                                                                | Issue                                                               | Fix                                                              |
| ------------------------------------------------------------------- | ------------------------------------------------------------------- | ---------------------------------------------------------------- |
| `Sandbox.Engine/Resources/Surface.cs`                               | `.FirstOrDefault().Value.ResourcePath` can NRE when no match exists | Added null check on match + warning log                          |
| `Sandbox.Engine/Services/Packages/PackageManager/PackageManager.cs` | `.First()` throws if no package found                               | Switched to `.FirstOrDefault()` (consistent with other overload) |
| `Sandbox.Engine/Scene/GameObjectSystems/ClutterGridSystem.cs`       | `_subscribedTerrains.First()` unsafe if list changes during runtime | Replaced with safe lookup + guard                                |
| `game/addons/tools/Code/Editor/AssetPicker/GenericPicker.cs`        | `.FirstOrDefault().Package` can still NRE if nothing selected       | Split into variable + null-safe access                           |
| `Sandbox.Engine/Scene/Components/Game/BaseChair.cs`                 | assumes exit points always exist                                    | filter valid points + fallback to seat position                  |
| `Sandbox.Engine/Game/Mount/MountUtility.cs`                         | `.First()` used even though guarded by count check                  | replaced with `.FirstOrDefault()` for consistency                |
| `Sandbox.Compiling.Test/Tests/BlacklistTest.cs`                     | test assumes indexing always safe if earlier assertion passes       | switched to explicit index access for clearer failure            |

> **Note:** `NetworkConsoleCommands.FindAndJoinLobby()` was initially modified but reverted — `Lobby` is a struct so `FirstOrDefault()` returns a default struct, not null, and the existing `.Any()` guard already makes `.First()` safe there.

---

N/A

---

## Checklist

* [x] Public APIs are documented (if applicable)
* [x] Unit tests added where applicable and all passing

### DIFF
--- a/engine/Sandbox.Compiling.Test/Tests/BlacklistTest.cs
+++ b/engine/Sandbox.Compiling.Test/Tests/BlacklistTest.cs
@@ -1,4 +1,4 @@
-﻿using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using System;
 using System.Collections.Generic;
@@ -201,7 +201,7 @@ public partial class BlacklistTest
 
 		// Prohibited type 'System.Runtime.CompilerServices.Unsafe.As<T>(object)' used
 		Assert.AreEqual( 1, diagnostics.Count );
-		Assert.AreEqual( "SB500", diagnostics.FirstOrDefault().Id );
+		Assert.AreEqual( "SB500", diagnostics[0].Id );
 	}
 
 	[TestMethod]
@@ -218,7 +218,7 @@ public partial class BlacklistTest
 
 		// (1,15): error SB500: Prohibited type 'System.Runtime.CompilerServices.Unsafe.As' used
 		Assert.AreEqual( 1, diagnostics.Count );
-		Assert.AreEqual( "SB500", diagnostics.FirstOrDefault().Id );
+		Assert.AreEqual( "SB500", diagnostics[0].Id );
 	}
 
 	[TestMethod]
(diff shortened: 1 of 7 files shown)


lolleko:
Hey, thanks for this contribution but we don't really care about NREs in tests, in most cases we would actually want the tests to fail if something becomes null.

For the rest of the "fixes", as long as they are not actually causing errors, i would also keep everything as it is.
By default, it's usually better not to silently catch errors, so you know when something goes wrong.

PR #11050, opened 2026-05-26
Fix TriggerHurt's damage timer

Deotexh:
# Pull Request

---

## Summary

TriggerHurt timer is reset everytime it goes past the rate, even if no one got hurt
The PR makes it so the timer is only resetted when something got damaged

## Motivation & Context

Because in some games using a Trigger Hurt with high rate, this makes possible scenarios where:
With a rate of 45 seconds for example, the player goes inside a TriggerHurt box when the timer got set at 45.1f (so the timer got reset 0.1 second before the player enter) and the player has 45 seconds of "invincibility window", even though they have not taken damage (which would only then grant them this window)

Fixes: 

## Implementation Details

I put the logic out of the for loop, so that the first IDamageable in the loop doesn't deny damage to other IDamageable by setting the timer to 0

## Checklist


Deotexh:
Wait, I cannot declare a variable inside a function?

PR #11051, opened 2026-05-27
Prop: scatter gibs from explosion origin when killed by explosion

Xenthio:
## Problem

Stationary props killed by an explosion spawn gibs in place with no outward velocity. `CreateGibs()` inherits velocity from the prop's own `Rigidbody.PreVelocity`, which is zero if the prop wasn't moving before the explosion hit.

## Fix

- Added `LastExplosionOrigin` property to `Prop` — set in `OnDamage` whenever the damage has the `"explosion"` tag.
- In `CreateGibs()`, if `LastExplosionOrigin` is set, apply an outward impulse to each gib's rigidbody proportional to the gib's mass, with a linear distance falloff (full at origin, zero at 512 uu), plus a small torque so pieces spin as they fly.

This stacks with the existing `PreVelocity` inheritance — moving props killed mid-air still get their own velocity, and then also get the explosion scatter on top.

No new public API surface required (the property is non-Sync and purely internal to the break path).

### DIFF
--- a/engine/Sandbox.Engine/Scene/Components/Game/Prop.cs
+++ b/engine/Sandbox.Engine/Scene/Components/Game/Prop.cs
@@ -362,7 +362,7 @@ public class Prop : Component, Component.ExecuteInEditor, Component.IDamageable
 		if ( IsExplosive && damage.Tags.Contains( "impact" ) )
 		{
 			Health = 0;
-			Kill();
+			Kill( damage );
 			return;
 		}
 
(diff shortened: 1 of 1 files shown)


Xenthio:
Note: this is a targeted fix for the most common case (explosion kills stationary prop). The underlying issue is broader — `CreateGibs()` reads `PreVelocity` which is captured at frame start, so any impulse applied this frame (bullet, explosion, physgun impact) isn't integrated yet and won't be reflected in gib velocity.

A fuller fix would thread the killing `DamageInfo` through the call chain: `Kill(DamageInfo)` -> `OnBreak(DamageInfo)` -> `CreateGibs(DamageInfo)`, then apply the damage force directly to gibs from there. That would handle all damage types uniformly and remove the need for `LastExplosionOrigin` entirely. Happy to extend the PR if that direction is preferred.

PR #11053, opened 2026-05-27
DamageInfo: add Vector3? Force property; use it to push gibs on prop break

Xenthio:
## What

Adds `Vector3 Force` to `DamageInfo` (same idea as HL2's `CTakeDamageInfo::m_vecDamageForce`) and fixes gibs spawning with no velocity when a prop is killed.

## The bug

`CreateGibs` inherits velocity from `rb.PreVelocity`, which is captured at the start of the physics tick. Any impulse applied that same tick (bullet, explosion) hasn't been integrated yet so it's not in `PreVelocity` and gibs get nothing. Impact kills (collision damage) have the opposite problem, rb.Velocity has already been drained by the collision solver before gibs even spawn.

## Fix

Two things in `CreateGibs`:
- For **impact** kills (`damage.Tags.Contains("impact")`), use `rb.PreVelocity` - the pre-collision snapshot is what we want.
- For **everything else**, use `rb.Velocity` - which now reflects impulses applied this tick by `Prop.OnDamage` or `RadiusDamage`.

`Prop.OnDamage` also applies `damage.Force` as an impulse to its own prop

`RadiusDamage` sets `damage.Force` per-target in the damageables loop so explosions work without any extra code on the caller side.

force damage stuff like many games already use and expect

## Notes

`Force` is a plain `Vector3`, zero means no push. Convention is direction x magnitude, not mass-scaled; consumers multiply by their own mass when applying as an impulse.

Companion sandbox PR: https://github.com/Facepunch/sandbox/pull/273

Videos in the s&box discord because they're too big to upload here:
https://discord.com/channels/833983068468936704/833983416390385685/1509092234564337694

### DIFF
--- a/engine/Sandbox.Engine/Scene/Components/Game/Prop.cs
+++ b/engine/Sandbox.Engine/Scene/Components/Game/Prop.cs
@@ -362,7 +362,7 @@ public class Prop : Component, Component.ExecuteInEditor, Component.IDamageable
 		if ( IsExplosive && damage.Tags.Contains( "impact" ) )
 		{
 			Health = 0;
-			Kill();
+			Kill( damage );
 			return;
 		}
 
@@ -386,7 +386,7 @@ public class Prop : Component, Component.ExecuteInEditor, Component.IDamageable
 
 		if ( Health <= 0 )
 		{
-			Kill();
+			Kill( damage );
 			Health = 0;
 		}
 	}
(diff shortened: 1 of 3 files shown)


Xenthio:
the pr auto tests check is failing for reasons unrelated to this pr btw

Xenthio:
I've considered maybe the impact bullet force should live in Prop? but I'm unsure. Will leave it as is unless asked.

Xenthio:
I have an alternative implementation, it fixes the same bug, loses the ability for individual directions on gibs a bit, also plays nice with pre existing gamemodes. 
Engine side changes: https://github.com/Xenthio/sbox-public/commit/c28f87740db2c44f4a67b02a66c210ada8ec16ff

Sandbox side changes: https://github.com/Xenthio/sandbox/commit/afaccd4e11b8b0738bb8d05f2e08f07aa84ed480

But i scuffed the commit up and included some other stuff  (Ignore OnGibsCreated, FindGibAtPoint)

But if that approach looks reasonable i can move it into here. (does move prop impact force into prop.cs so its ran in a certain order, tho)

Xenthio:
better implementation done

PR #11080, opened 2026-05-30
Log error when font load fails and falls back to system font

Mykol-Word:
## Summary

Displays a brief error message when loading a font fails and falls back to system font. 

## Motivation & Context

Previously loading a font would fail silently and fall back to the system font. This made using custom fonts confusing since it was unclear if the custom font was failing to load or wasn't even being loaded at all. Users may also just not even notice their font failed to load.

This addition was made in response to an issue mentioned on discord.

## Screenshots

image

## Checklist


garrynewman:
I would support a Log.Warning

Mykol-Word:
Sure, I don't really care as long as its obvious.

PR #11143, opened 2026-06-05
Copy bounds of Model Renderer onto newly initialised Box Collider component

ElmarTalibzade:
# Pull Request

---

## Summary

When added, the Box Collider will now assume the shape of the Model Renderer bounds.

## Motivation & Context

I found myself deleting Model Collider in favour of a simpler, more performant Box Collider as my game did not require elaborate hitboxes (especially for static environments).

I've noticed that Box Collider (unlike in Unity), does not react to whether the GameObject has a Model Renderer, which has clear bounds, and always defaults to standard 50x50x50 box. This makes it a pain having to match the shape of the model by hand.

https://github.com/user-attachments/assets/e626b08f-b678-4993-8460-88ef6e0e5a7f

## Implementation Details

I've added `OnAwake()` override that does the following:
- Checks whether box center and size are default. 
- If they are then look for a component that implements `IHasBounds`
- If there isn't, then do nothing (use the usual size/center properties)
- If there is, then set the size/center to match the bounds of said component

It's not pretty, and I don't expect this PR to be easily approved, but I haven't found any better mechanism without having to resort to bigger refactoring.

## Checklist


aylaylay:
This is awful, the naming is awful, relying on it being a value of 50 is awful, doing it in OnAwake is awful

If we're going to do this, I'm going to make it an editor only thing when it only does it when adding a new component

ElmarTalibzade:
Thanks for reviewing my PR, @aylaylay

I agree with most of your statements. but this one needs addressing:

> doing it in OnAwake is awful

You're already doing this:

`ModelCollider` component has an `OnAwake` method that auto-assigns a model if there's a neighbouring `ModelRenderer` component ([source](https://github.com/Facepunch/sbox-public/blob/master/engine/Sandbox.Engine/Scene/Components/Collider/ModelCollider.cs#L26-L38)). 

I can't think of a scenario where this is useful at runtime, but feel free to correct me.

PR #11193, opened 2026-06-10
Add local loopback socket for multi-instance testing

### LINKED ISSUE #10965: Allow multiple local client instances to join the same dedicated server
### What can't you do?

I'm unable to connect two local client instances to the same dedicated server.

When the second instance joins, the first one gets disconnected with an error message, and the second one connects instead.

I want multiple local client instances to be able to join the same dedicated server at the same time.

The best approach would likely be to reuse the same fake SteamID system used by the editor when using Play and then Join with new instance.

A launch option could also be added to enable this behavior for standalone clients, for example:

`+allowInsecure`

### What have you tried?

I tried starting a dedicated server, joining it with one client, then launching a second client and joining the same server.

The first client gets disconnected, and only the second client stays connected.

### Additional context

This is needed for local dedicated server testing with multiple clients on the same machine.

dictateurfou:
# Pull Request

## Summary

Adds an optional local loopback socket for dedicated servers.

When enabled, dedicated servers expose a localhost TCP transport, allowing multiple game instances running on the same machine to connect to the server. This brings dedicated server testing closer to the existing local multiplayer workflow available in standalone/editor sessions.

Clients can connect through the loopback transport using `connect local`.

---

## Motivation & Context

Testing multiplayer features against a dedicated server often requires running multiple game instances locally.

While standalone and editor sessions already expose a local TCP transport for local connections, dedicated servers do not currently offer an equivalent option. This makes local multi-instance testing more difficult and typically requires additional machines or external networking.

This change introduces an optional localhost transport that can be enabled when needed for development and testing.

Fixes: #10965

https://github.com/Facepunch/sbox-public/issues/10965
---

## Implementation Details

* Added a new networking convar: `net_allow_local`.
* When enabled, `CreateDedicatedServer()` adds an additional `TcpSocket` bound to `127.0.0.1:55333`.
* Clients can connect to the dedicated server through the loopback transport using `connect local`.
* The feature is opt-in and disabled by default.
* Existing dedicated server behavior remains unchanged unless explicitly enabled.
* Reuses the existing `TcpSocket` implementation already used for local connections in standalone/editor environments.

---

multi-instance

---

## Checklist


engine/Sandbox.Engine/Systems/Networking/Networking.cs
```diff
@@ -521,6 +524,11 @@ public static void CreateLobby()
 			net.AddSocket( DedicatedServer.IpSocket );
 			net.AddSocket( DedicatedServer.IdSocket );
 
+			if ( AllowLocal )
+			{
+				net.AddSocket( new TcpSocket( "127.0.0.1", 55333 ) );
```
lolleko, line 529: Port should be configurable

lolleko (changes requested):

engine/Sandbox.Engine/Systems/Networking/Networking.cs
lolleko: actually nvm its probably fine

dictateurfou: Yes, I can do it if you want. I used this port by default because it was the same for local P2P instances, and it allows using `connect local` instead of the IP.

I also don’t know if it would work with another port, because currently with `net_hide_address` set to `0`, we cant connect using the IP with multiple instance.

Let me know if it’s necessary. I can add the port, that’s not a problem. I assume there is already a parameter for that somewhere ?

PR #11194, opened 2026-06-10
Max Payne 2 mount

UnlitSpirit:
## Summary

 Adds a mount for **Max Payne 2: The Fall of Max Payne** (Steam app `12150`), structured to mirror the Quake mount (`Sandbox.Mounting.Quake`): a `BaseGameMount` that indexes the game's
  archives, with `ResourceLoader<T>` / `SceneLoader<T>` implementations per resource type, and mount shaders under `game/mount/maxpayne2/Assets/shaders`.

  What it surfaces:

  | Resource | Source | Notes |
  |---|---|---|
  | Scene | `.ldb` levels | static world per room, lightmaps, baked ambient lighting, collision |
  | Model | `.kf2` | static props and fully skinned characters with the game's animation library as model sequences |
  | Texture | `.dds .tga .pcx .jpg` | TGA normalized to top-down, PCX/JPG via Bitmap |
  | Sound | `.wav` | most of the game's audio is MS-ADPCM, decoded to PCM16 |

  ## Motivation & Context

- I love max payne

  Fixes: N/A

  ## Implementation Details

- **RAS archives**: the container is XOR-enciphered (header, file table and folder table each encrypted independently with a per-archive seed) with per-file LZSS compression. `Ras.cs`
  decrypts and inflates on demand.
- **KF2 models**: tagged-value stream with chunked layout. Chunk sizes include their own 13-byte header; the nested NODE chunk reports an unreliable size, so it is parsed field by
  field. Mesh node transforms are baked into vertices before coordinate conversion.
- **Skeletal characters** are a four-file system (mesh `.kf2` + skin `.skd` + shared per-sex skeleton/pose + an animation library of ~350 clips). The loader stitches these together and
  bakes every library clip as a named model sequence. Notable format quirks handled:
    - Bones are passed to `ModelBuilder` in model space with parent-local frames to `AnimationBuilder`, matching what the Quake/GoldSrc mounts do.
    - Every female skeleton has one mirrored bone (`Bip01 Breast-R`, a det = -1 mirror clone). The matrix-to-quaternion conversion rebuilds an orthonormal right-handed basis first,
  otherwise that bone garbles into a non-unit quaternion.
    - Animation tracks with an empty parent name are local to their skeleton parent, not the root (facial tracks rely on this).
    - Facial animation uses matrix scale (blinks and lip movement squash helper bones); frames carry per-axis scale relative to the bind so unkeyed bones stay identity.
    - One skin (`MaxPayne.skd`) names bones with different casing than the skeleton file; bone names are normalized to the skeleton's casing.
  - **LDB2 levels**: rooms hold the full static world. Lightmaps render through `mp2_world.shader` (same pattern as `quake_generic`: diffuse from uv0, lightmap atlas from uv1). MP2 has no  dynamic lights; lighting is baked lightmaps plus per-room volume-light grids, which are reduced to shadowless point lights as an ambient approximation. Render geometry is also used for
  collision.
  - **Coordinate conversion**: MP2 is meters, Y-up, opposite handedness. Positions map `(x, y, z) -> (-x, -z, y)` with a winding flip, scaled to inches.
  - **Meshless `.kf2` files** (animation libraries, camera paths, texture-set variants) are filtered at registration by sniffing top-level chunks for mesh data, so the asset browser only
  lists real models.
  - **Sounds**: `SoundFile.FromWav` passes ADPCM through as raw PCM, so format tag 2 files are decoded to PCM16 first and loaded with `FromPcm`.

image
image
image

https://github.com/user-attachments/assets/5f67c566-75be-4538-a683-f244516da57e

  ## Checklist

### DIFF
--- a/engine/Mounting/Sandbox.Mounting.MaxPayne2/Resource/MaxPayne2Model.cs
+++ b/engine/Mounting/Sandbox.Mounting.MaxPayne2/Resource/MaxPayne2Model.cs
@@ -23,7 +23,6 @@ class MaxPayne2Model( string fileName ) : ResourceLoader<MaxPayne2Mount>
 		var dir = System.IO.Path.GetDirectoryName( FileName )?.Replace( '\\', '/' );
 		var searchDirs = BuildSearchDirs( dir, kf2.TextureDirs );
 		var builder = Model.Builder.WithName( Path );
-		var textureCache = new Dictionary<string, Texture>( StringComparer.OrdinalIgnoreCase );
 
 		var rig = LoadRig( dir );
 		var animatedNodes = rig is null && kf2.Animations.Count > 0 && kf2.Nodes.Count > 0;
@@ -44,10 +43,10 @@ class MaxPayne2Model( string fileName ) : ResourceLoader<MaxPayne2Mount>
 			var indices = BuildIndices( kf2, prim );
 
 			var mesh = rig is not null
-				? BuildRigMesh( kf2, prim, positions, indices, searchDirs, textureCache, rig )
+				? BuildRigMesh( kf2, prim, positions, indices, searchDirs, rig )
 				: animatedNodes
-					? BuildSkinnedMesh( kf2, prim, positions, indices, searchDirs, textureCache, boneIndexByNode )
-					: BuildMesh( kf2, prim, positions, indices, searchDirs, textureCache );
+					? BuildSkinnedMesh( kf2, prim, positions, indices, searchDirs, boneIndexByNode )
+					: BuildMesh( kf2, prim, positions, indices, searchDirs );
 
 			var baseIndex = collisionVerts.Count;
 			collisionVerts.AddRange( positions );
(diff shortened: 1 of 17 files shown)


engine/Mounting/Sandbox.Mounting.MaxPayne2/Resource/MaxPayne2Model.cs
```diff
@@ -0,0 +1,670 @@
...
+		foreach ( var dir in searchDirs )
+		{
+			foreach ( var ext in new[] { ".dds", ".tga", ".jpg", ".pcx" } )
+			{
+				var candidate = string.IsNullOrEmpty( dir ) ? name + ext : $"{dir}/{name}{ext}";
+				var data = Host.GetFileBytes( candidate );
+				if ( data is not null )
+					return MaxPayneImage.Load( data );
+			}
+		}
```
aylaylay, line 768: does this need to be loading the texture again for every model? you registered textures in the mount, can request them

aylaylay (changes requested):

aylaylay:
I think max payne 2 has ragdolls, do you support them?

engine/Mounting/Sandbox.Mounting.MaxPayne2/Resource/MaxPayne2Model.cs
UnlitSpirit: Thank you for taking the time to review the PR, I will look into this

UnlitSpirit: Should now hopefully use the proper system to load registered textures

UnlitSpirit:
Added ragdoll support

https://github.com/user-attachments/assets/efb795a1-9053-4e21-964a-4ffa10b412d6

Issue #10492, opened 2026-04-15
Scenes not loading (can't be found) occasionally, restarting fixes


KenneyNL:
### Version / Commit

26-04-15

### Describe the bug

Often scenes will fail to load when using `Game.ActiveScene.LoadFromFile("scenes/menu.scene");`, when it fails it'll act like there's no scene file in the location. Restarting the editor fixes the problem.

### To Reproduce

1. Create two scenes
2. From one scene in a razor @code load "Game.ActiveScene.LoadFromFile("scenes/second.scene")" _(this is what I've tested, it could also be broken outside of razor files)_
3. Occasionally it'll fail, I can't pinpoint when it will start failing but it will keep failing until an editor restart

### Expected behavior

The expected result would be for it to find and load the scene file, always

DoctorGurke:
I get this with system scene and when loading scenes manually. Sometimes it also seems to get fixed by manually re-saving the scene file (even with no unsaved changes). When the system scene stops loading once again I have to go into project settings and resave those sometimes...

FoxOnALeash:
Same here. 
A workaround is re-compiling the scene or just re saving it.

DoctorGurke:
AAAAAH 

<img width="454" height="43" alt="Image" src="https://github.com/user-attachments/assets/d1dbc019-3c79-44e1-bd7c-63bac93b5511" />

PR #11242, opened 2026-06-16
Fix "No valid Scene was found in SceneLoadOptions"

766974616c79:
## Summary

In order for `SceneLoadOptions` to find a scene, it uses a hashmap that maps the resource path to the resource. This hashmap is updated when a `GameResource` instance is created or destroyed. For destruction, it relies on the garbage collector. However, before the garbage collector runs, it is possible for another `GameResource` with the same path to be registered. Then, once the garbage collector executes, the cache removes that mapping, causing the scene to become unfindable.

The issue is that every time the scene is saved, a new instance of `SceneFile` is created. And `SceneFile` is a `GameResource`.

Fixes: #10130 #10492 

## Implementation Details

I avoid initializing a `SceneFile` as much as possible when saving.

## Checklist


trende2001:
I need thisss

lolleko:
This might be fixed on staging, could you test if this is still needed?

PR #11245, opened 2026-06-16
Fix Asset Browser locations tree freezing when a pinned folder is missing

trende2001:
## Summary

Fixes the Local Asset Browser's locations tree becoming unresponsive to clicks when a pinned folder no longer exists on disk.

## Motivation & Context

`LocalAssetLocations.OnFrame()` short-circuits when the pin list is unchanged, guarded by `PinsNode.Children.Count() == Pins.Count()`. But the rebuild loop skips pins whose folder is missing (`if ( !di.Exists ) continue;`), so once a pin points at a deleted/renamed folder the counts never match and `Rebuild()` runs every frame.

Each `Rebuild()` replaces the tree's `VirtualWidget` instances, and `BaseItemWidget` only registers a click when the press-time `pressedItem` still matches `hovered` at release. The per-frame rebuild orphans that widget mid-click, so clicks get dropped. (`OnFrame` calls `Rebuild()` directly, bypassing the `pressedItem` guard in `UpdateIfDirty`.)

## Implementation Details
Guard on a content snapshot of the pin list (`SequenceEqual`) instead of the child count. `Pins` is `internal static` and shared across instances, so this still picks up changes made elsewhere, but no longer mismatches forever when a folder is missing.

N/A

## Repro

Pin a folder in the Local asset browser, delete or rename it on disk, then click items in the locations tree.

## Checklist

### DIFF
--- a/game/addons/tools/Code/Editor/AssetBrowser/Local/LocalAssetLocations.cs
+++ b/game/addons/tools/Code/Editor/AssetBrowser/Local/LocalAssetLocations.cs
@@ -157,10 +157,11 @@ public class LocalAssetLocations : AssetLocations
 	[EditorEvent.Frame]
 	public void OnFrame()
 	{
-		if ( PinsNode?.Children.Count() == Pins.Count() && !_pinsDirty )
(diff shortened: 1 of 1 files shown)


solwllms:
I've pushed a slightly different solution that should fix this, but thanks for taking a look

PR #11306, opened 2026-06-24
[Scene Mesh] Convert Selection Toggle

internetfishy:
# Pull Request

---

## Summary

Adds a button in selection options to always transfer selection. This was previously done by holding alt when changing tools.

Holding alt will now invert the current setting.

## Motivation & Context

I don't think many people know it exists and is hidden behind hotkeys. It is a very useful selection option that I think is worth having a toggle for. It is on by default in other modelling programs like Blender.

Fixes: N/A

## Implementation Details

Adds a toggle button. Off by default to maintain current tool behavior.

convert

## Checklist


CarsonKompon:
This is nice, but it should probably use `EditorPreferences` just like the other options (like `EditorPreferences.BackfaceSelection`)

PR #11329, opened 2026-06-25
Reorder dedicated server NetworkSystem initialization to ensure Networking.IsActive is set by OnStart

### LINKED ISSUE #11170: `Networking.IsActive` is false on a dedicated server
### Version / Commit

26.06.03

### Describe the bug

I have a component where `OnStart` if `!Networking.IsActive` is false it does some stuff (perhaps i'm in singleplayer mode.)

When I launch a dedicated server, this code still runs because `Networking.IsActive` is returning false.

### To Reproduce

1. Use `Networking.IsActive` on a dedicated server, ideally in a `OnStart` in a component

### Expected behavior

It should return true.
A dedicated server saying networking is not active is silly. It's a server, of course networking is active.

### Media/Files

_No response_

### Additional context

_No response_

deadmonstor:
# Pull Request

---

## Summary

Changing the ordering of how dedicated servers create NetworkSystem so that Networking.IsActive is true in OnStart

Fixes #11170 

## Implementation Details
Currently on lobbies we create the network system and then init everything which allows Networking.IsActive to be set when OnStart is called, however dedicated servers do it the other way around meaning we need to assign the network system beforehand and cleanup if it fails. 

https://github.com/Facepunch/sbox-public/blob/8c9e8cbf53b1e68050636a972959a6ebced7e35c/engine/Sandbox.Engine/Systems/Networking/Networking.cs#L543-L575

With change:
image

Without change:
image

Code just incase you guys want to check
```csharp
using Sandbox;

public sealed class NetworkingIsActiveDedicatedServer : Component
{
	protected override void OnStart()
	{
		base.OnStart();

		Log.Info($"NetworkingIsActiveDedicatedServer: {Networking.IsActive}");

		Log.Info("Waiting for 5 seconds to check if we're running as a dedicated server...");
		Task.Delay(5000).ContinueWith(_ =>
		{
			if (Networking.IsActive)
			{
				Log.Info("Networking is active, we are running as a dedicated server.");
			}
			else
			{
				Log.Info("Networking is not active, we are NOT running as a dedicated server.");
			}
		});
	}
}
```

## Checklist

### DIFF
--- a/engine/Sandbox.Engine/Systems/Networking/Networking.cs
+++ b/engine/Sandbox.Engine/Systems/Networking/Networking.cs
@@ -505,9 +505,6 @@ public static partial class Networking
 
 	static async Task<bool> CreateDedicatedServer( LobbyConfig config, CancellationToken token = default )
 	{
-		var success = await DedicatedServer.Start( config );
-		if ( !success ) return false;
-
 		lock ( NetworkThreadLock )
 		{
 			var net = new NetworkSystem( "server", Engine.IGameInstanceDll.Current.TypeLibrary )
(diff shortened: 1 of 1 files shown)


Remscar:
Wow I encountered this issue a few weeks ago and it was really frustrating. Thanks for fixing this, I hope Facepunch merges it!

PR #11332, opened 2026-06-26
Enable Library references and interoperability

obselate:
# Pull Request

---

## Summary

This PR enables libraries to reference other libraries and resolve references topologically

## Motivation & Context

[ongoing discussion about library dependencies](https://github.com/Facepunch/sbox-public/issues/4322#issuecomment-4805954144)

## Implementation Details

This implementation is a rudimentary version of library <-> library interoperability, allowing a library author to declare references manually in their library's project settings. Enumeration of packages is type-agnostic so minimal changes were needed.

Libraries are topologically sorted and any cyclic dependencies are logged by `SortLibrariesForCompilation` so this was already production-ready but gated by the allow-list simply not including libraries.

Version resolution and "dependency hell" is not a concern in this PR. s&box already stops short of version resolution which is the exact same path taken here. As mentioned in the discussion in #4322 - it is not necessarily FacePunch's job to solve the dependency problem in the same way npm does, but a middle-ground that allows the community of library creators to create and maintain their own dependency models while the Library Manager simply hosts those libraries is a workable solution with minimal effort applied. 

This change enables users who want to utilize more generalized libraries, reduces code duplication, and allows for composition and a quicker / easier game development cycle.

See visuals and further explanation in the #4322 issue post linked above.

## Checklist

- [ x ] Code follows existing style and conventions
- [ x ] No unnecessary formatting or unrelated changes
- [ x ] Public APIs are documented (if applicable)
- [ x ] Unit tests added where applicable and all passing
- [ x ] I’m okay with this PR being rejected or requested to change 🙂

### DIFF
--- a/engine/Sandbox.Engine/Services/Packages/PackageManager/PackageLoader.cs
+++ b/engine/Sandbox.Engine/Services/Packages/PackageManager/PackageLoader.cs
@@ -181,6 +181,9 @@ internal sealed partial class PackageLoader : IDisposable
 					if ( baseHotloaded ) return true;
 					break;
 
+				case "library":
+					break;
+
 				default:
 					return false;
 			}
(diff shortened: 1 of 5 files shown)


lolleko:
Thanks for the contribution, but we have something else planned for libraries.

PR #11338, opened 2026-06-27
Clutter fix publish

PolEpie:
# Pull Request

---

## Summary

The clutter paint was working perfectly on normal map, but once published and use my MapInstance, you'd love every painting you made

## Motivation & Context

My mapper was doing great job on his addon that target my project and after publish i wanted to try load the map on my project via workshop, but it looked empty

Fixes: 

## Implementation Details

So i dig a little and at first thought well maybe the models don't get referenced on published, so i see that since it used BinaryBlob, the scanner of reference would not see it and so none models would be published if used exclusivly on clutter painting
So i did IBlobReference that let us specify what we reference

But it wasnt enough, it did fix publish of models but still not spawning a single model, so i dig a bit and found out the _d data of the clutter embeded into the scene_c wasnt read or was empty, a tired dig with claude and found a workaround to pull out these data and apply it on MapInstance

So then a single tree got placed, it was one that also was placed manually into the scene, so already referenced
The conclusion was that ClutterGridSystem, loaded wrong the model, i found out that switch to Model.Load instead of ResourceLibrary.Get<Model> fixed it

So now clutter paint work perfectly !

image

## Checklist

### DIFF
--- a/engine/Sandbox.Engine/Scene/GameObjectSystems/ClutterGridSystem.cs
+++ b/engine/Sandbox.Engine/Scene/GameObjectSystems/ClutterGridSystem.cs
@@ -34,10 +34,10 @@ public sealed partial class ClutterGridSystem : GameObjectSystem
 	[Property, Hide]
 	public ClutterStorage Storage
 	{
-		get => _storage;
+		get => field;
 		set
 		{
-			_storage = value;
+			field = value;
 			// The data may have been replaced wholesale (e.g. by a map load applying this
 			// scene's GameObjectSystem overrides). Mark dirty so the painted layer rebuilds
 			// on the next update.
--- a/engine/Sandbox.Engine/Scene/GameObjectSystems/ClutterGridSystem.cs
+++ b/engine/Sandbox.Engine/Scene/GameObjectSystems/ClutterGridSystem.cs
@@ -25,12 +25,25 @@ public sealed partial class ClutterGridSystem : GameObjectSystem
 	private readonly List<Terrain> _sceneTerrains = [];
 	private readonly HashSet<ClutterLayer> _layersToRebuild = [];
 
+	private ClutterStorage _storage = new();
+
 	/// <summary>
 	/// Storage for painted clutter model instances.
 	/// Serialized with the scene - this is the source of truth for painted clutter.
 	/// </summary>
 	[Property, Hide]
-	public ClutterStorage Storage { get; set; } = new();
+	public ClutterStorage Storage
+	{
+		get => _storage;
+		set
+		{
+			_storage = value;
+			// The data may have been replaced wholesale (e.g. by a map load applying this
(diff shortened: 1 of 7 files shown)


engine/Sandbox.Engine/Scene/GameObjectSystems/ClutterGridSystem.cs
```diff
@@ -19,12 +19,25 @@ public sealed partial class ClutterGridSystem : GameObjectSystem
...
 	/// Serialized with the scene - this is the source of truth for painted clutter.
 	/// </summary>
 	[Property, Hide]
-	public ClutterStorage Storage { get; set; } = new();
+	public ClutterStorage Storage
+	{
+		get => _storage;
+		set
+		{
+			_storage = value;
```
antopilo: use field keyword instead

antopilo (changes requested):
Overall needs a good clean up, seems to be duplicated code in the mapinstance load binary blob, it should just work without doing manual stuff like that.

Trim it down as much as you can, we dont need wall of texts everywhere

PolEpie:
Okay i did a 2nd pass, so now MapInstance just need BlobDataSerializer.LoadFrom and then into it i managed to make it load from PackageManager.MountedFileSystem fs else the _d wouldnt be loaded, maybe it's still not the best way of doing it, i'm looking for your opinion here

I also trim down every comment and used field

Thanks for your review now i see what to improve into my future PRs
Also with latest commit that does ApplyGameObjectSystemOverrides being able to unload the GameObjectSystem values
But it's a bit more long, i can split it if needed, it was simply a issue i wanted to fix and i saw so many issues related x) but it was a good journey

Ethraxxx:
Hey, do you know when this PR will be merged?

ErazorWhite:
We also need this fix.

The editor stores mesh geometry in a separate .scene_d file and keeps only `$blob` references in the scene itself. When a map is loaded from a published package the engine never reads that data - every mesh comes out empty, zero errors, while skybox/lights/models still work, so it looks like a broken map rather than an engine bug.

I worked around it by writing the mesh data back into the `.scene` directly: a small script takes each blob from `.scene_d` and replaces the `$blob` refs with inline base64, which the engine still reads fine. Republish without saving the scene in the editor (pick "reload from disk" if it asks about external changes) - my 332-mesh map renders again. But every editor save converts the scene back to the broken format, so the conversion has to be redone before each publish until it's fixed properly

image

ErazorWhite:
@PolEpie try to re-request review from [antopilo](https://github.com/antopilo)
image

handsomematt:
@PolEpie can you rebase on latest and fix any conflicts, I can't triage otherwise

PolEpie:
@handsomematt done !
(Also please take a look at #11344 cause it's the 2nd fix for clutter to work with MapInstance)

PR #11359, opened 2026-06-30
Fix sync vars getting stuck on the client due to dropped/delayed ACKs

deadmonstor:
# Pull Request

---

## Summary

Currently sync vars do not check if the data in _predictedData is actually the right one they are acking, so its possible that we remove the wrong predictive data from an older snapshot id. This prevents a delayed or out-of-order ACK from removing a more recent predicted sync value, which could cause a synced variable to become "stuck" (incorrectly revert or stop being resent if they never receive it).

Examples:

• predicted=2, incoming=3 -> IsNewer(2,3) = false -> remove (OK)
• predicted=5, incoming=3 -> IsNewer(5,3) = true -> don't remove (keep prediction)

Fixes: #795 

## Implementation Details

TryGetHash now treats an expired prediction as a miss instead of falling back to older stored data. This forces a resend when the receiver likely didn’t get the predicted value, avoiding a synced variable becoming stuck on an older value in the case that we have "flickering" variables between two values. 

Server:
Value A, Value B, Value A (Server still thinks Value A is the value so no send), Value B

Client:
Value A, Value B (Ack fails to get in within time), Value B

Before: 

https://youtu.be/i1XP1RmZJGQ

After:

https://youtu.be/MXKmexLvS80

The project for this bug is in #795 

## Checklist

### DIFF
--- a/engine/Sandbox.Engine/Scene/Networking/DeltaSnapshots/DeltaSnapshotSystem.cs
+++ b/engine/Sandbox.Engine/Scene/Networking/DeltaSnapshots/DeltaSnapshotSystem.cs
@@ -237,7 +237,7 @@ internal class DeltaSnapshotSystem
 					dataToSend ??= SnapshotData.Pool.Rent();
 					dataToSend[slot] = value;
 
-					state.AddPredicted( entry, Time );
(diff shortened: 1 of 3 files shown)


Remscar:
You are a king among men

PR #11383, opened 2026-07-04
Joining snapshot now hides non always transmit objects if applicable

### LINKED ISSUE #10938: INetworkVisible / AlwaysTransmit: also apply to initial snapshot for fresh-joining clients
### What can't you do?

When AlwaysTransmit = false and a Component implements INetworkVisible, the engine correctly filters ongoing Sync and Transform updates per-connection. But the initial snapshot sent to a fresh-joining client appears to bypass the visibility callback — the new client receives every networked object's current WorldPosition and [Sync] values regardless of whether the visibility callback would have returned false for them.
This means a connecting player gets one free read of every networked entity's position at the moment of join. For competitive PvP this is an ESP/wallhack seed: continuous tracking is properly culled, but the initial frame leaks everything.
The docs hint at the current behavior — "Visibility determines whether the object receives ongoing network updates" — but there's no equivalent path for the initial snapshot.

When the engine builds the initial snapshot for a connecting client, for each object with AlwaysTransmit = false, call its root INetworkVisible.IsVisibleToConnection and skip the object if it returns false — same code path used for ongoing updates. The client wouldn't receive that object until they later become a valid recipient (which already works via the engine's existing visibility-change path).
If changing the default would break existing games, an opt-in flag on NetworkSpawnOptions would be enough

### What have you tried?

Verified the issue with a client-side debug overlay that reads WorldPosition from every Player proxy in the local scene — the only data a malicious client would have access to. After connecting to a host with AlwaysTransmit = false and INetworkVisible returning false for out-of-range connections, the overlay shows correct positions for distant entities at the moment of join. Continuous tracking afterward is correctly blocked (positions go [STALE] or the proxy disappears entirely), confirming that ongoing-update culling works as documented — only the initial snapshot leaks.
Workarounds considered: dropping [Sync] for transforms on relevant entities and broadcasting position via a custom chunk-filtered RPC tier. Doable but a substantial architectural change for what looks like a single engine-side fix.

### Additional context

INetworkVisible + AlwaysTransmit is a great API and works flawlessly for the post-join lifecycle. This request is just about closing the one remaining lifecycle stage. Thanks for reading.

deadmonstor:
# Pull Request

---

## Summary

This is fundamentally trying to fix a workflow problem: when you "hide" an object from a client, the expectation is that the client doesn't have the object at all, not that it still exists in their world but happens to be invisible. INetworkVisible should prevent the object from ever reaching clients for whom it's false, rather than leaking it to them via the initial snapshot packet.

## Motivation & Context

Fixes #10938 / Implements #10938

## Implementation Details

Overall this is how it should have been to begin with however if someone is relying on behavior of the object existing on the client regardless of if this is true or false might have a problem however I had a look at a few of the gamemode's and they aren't even using this feature so I think we are okay.

The other solution to this was just to drop all of the sync var's but I thought in general it just makes sense to not have objects on the client that aren't meant to be there, dependent on how its setup per gamemode it could actually reduce the amount of objects.

This does require tracking who knows about what object however that shouldn't be a problem as its all guid's anyways

Before: https://www.youtube.com/watch?v=2wtvSZDVpuc
After: https://www.youtube.com/watch?v=XQDwPhpbg70

```csharp
using Sandbox;
using static Sandbox.Component;

public sealed class TestNetworkVisible : Component, INetworkVisible
{
	public TestNetworkVisible()
	{
		Log.Info( "TestNetworkVisible created" );
	}

	protected override void OnAwake()
	{
		base.OnAwake();

		Log.Info( "TestNetworkVisible awake" );
	}

	[ConVar( "test_network_visible_debug" )]
	public static bool TestNetworkVisibleDebug { get; set; }

	public bool IsVisibleToConnection( Connection connection, in BBox worldBounds )
	{
		return TestNetworkVisibleDebug;
	}
}
```

The object that you put this on has to be network object + have AlwaysTransmit turned off. 

## Checklist

### DIFF
--- a/engine/Sandbox.Engine/Scene/Networking/NetworkObject.cs
+++ b/engine/Sandbox.Engine/Scene/Networking/NetworkObject.cs
@@ -392,6 +392,7 @@ internal sealed partial class NetworkObject : IValid, IDeltaSnapshot
 	internal readonly LocalSnapshotState LocalSnapshotState = new();
 
 	private readonly HashSet<Guid> _culledConnections = [];
+	private readonly HashSet<Guid> _createMessageConnections = [];
 	private readonly Dictionary<Guid, CullState> _cullStates = new();
 	private readonly SnapshotValueCache _snapshotCache = new();
 	private TimeUntil _nextUpdateCachedBounds;
@@ -410,6 +411,7 @@ internal sealed partial class NetworkObject : IValid, IDeltaSnapshot
 	{
 		LocalSnapshotState.RemoveConnection( id );
 		_culledConnections.Remove( id );
+		_createMessageConnections.Remove( id );
 		_cullStates.Remove( id );
 	}
 
@@ -428,6 +430,7 @@ internal sealed partial class NetworkObject : IValid, IDeltaSnapshot
 	private void ClearConnections()
 	{
 		LocalSnapshotState.ClearConnections();
+		_createMessageConnections.Clear();
 	}
 
 	bool IDeltaSnapshot.ShouldTransmit( Connection target )
@@ -446,6 +449,7 @@ internal sealed partial class NetworkObject : IValid, IDeltaSnapshot
 				if ( !_culledConnections.Remove( target.Id ) )
 					continue;
 
+				EnsureCreateMessageSent( target );
 				GameObject.Network.SetCullState( target, false );
 			}
 
@@ -496,6 +500,7 @@ internal sealed partial class NetworkObject : IValid, IDeltaSnapshot
 
 				LocalSnapshotState.RemoveConnection( target.Id );
 				GameObject.Network.SetCullState( target, false );
+				EnsureCreateMessageSent( target );
 
 				shouldTransmitToAny = true;
 				state.Culled = false;
@@ -519,11 +524,32 @@ internal sealed partial class NetworkObject : IValid, IDeltaSnapshot
 	}
 
 	/// <summary>
+	/// Ensure that a create message has been sent to the specified <see cref="Connection"/>. If it has not, then send it.
+	/// </summary>
+	/// <param name="target"></param>
+	internal void EnsureCreateMessageSent( Connection target )
+	{
+		if ( !_createMessageConnections.Add( target.Id ) )
+			return;
+
+		target.SendMessage( GetCreateMessage() );
+	}
+
+	/// <summary>
+	/// Mark that a create message has been sent to the specified <see cref="Connection"/>.
+	/// </summary>
+	/// <param name="target"></param>
+	internal void MarkCreateMessageSent( Connection target )
+	{
+		_createMessageConnections.Add( target.Id );
+	}
+
+	/// <summary>
 	/// Is this network object visible to the provided <see cref="Connection"/>. We'll check if we
 	/// have a culler component and use that, but we'll also use our bounds to determine if we're
 	/// visible.
 	/// </summary>
-	private bool IsVisible( Connection target, BBox worldBounds )
+	internal bool IsVisible( Connection target, BBox worldBounds )
 	{
 		// Do we have a INetworkVisible? We're going to let that take priority.
 		var go = GameObject;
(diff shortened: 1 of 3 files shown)


aidencurtis:
This is amazing

K3rhos:
Not sure if this is my problem, but you could have broke the feature (Not accusing you, just really suspect that this has been changed recently and I experience issue now using INetworkVisible/AlwaysTransmit related stuff while I had no issue before):

https://github.com/Facepunch/sbox-public/issues/11523

K3rhos:
## I quickly decided to put claude onto your PR just to get a quick overview (not pretending this is perfect, it's AI after all, always need to double check, but I guess this would give you a hint):

Here's the exact anatomy, with all three pieces on the table now. The dedup mechanism the PR introduced is one set per networked object:

The guard — NetworkObject.cs:395: _createMessageConnections, a per-object set of connections that already got a create message. EnsureCreateMessageSent (NetworkObject.cs:530) sends only if the connection isn't in that set:

```cs
internal void EnsureCreateMessageSent( Connection target )
{
    if ( !_createMessageConnections.Add( target.Id ) )
        return;
    target.SendMessage( GetCreateMessage() );   // ← the duplicate create comes from here
}
```
The whole scheme only works if every path that delivers a create also records it in that set. The PR wired that up in three sender paths (SceneNetworkSystem.cs:167, :197, :226 — the NetworkSpawn/broadcast paths) and in SendFullServerSnapshot (SceneNetworkSystem.cs:317-326, which walks the snapshot afterward and calls MarkCreateMessageSent(connection) for every included object).

The hole — SceneNetworkSystem.cs:412, the GetSnapshot override that serves a joining client:

```cs
public override void GetSnapshot( Connection source, ref SnapshotMsg msg )
{
    ...
    Game.ActiveScene.SerializeNetworkObjects( source, msg.NetworkObjects );  // line 430
    ...
}
```
SerializeNetworkObjects (in Scene.Network.cs) now filters per-connection: a transmit-off object is included when root.ShouldTransmit(source) || target.IsVisible(source, worldBounds). So your waterramp/basketball creates go into the join snapshot — but nothing in this path ever calls MarkCreateMessageSent(source). The connection got the objects, and the object's _createMessageConnections set doesn't know it. That's the missing bookkeeping, and it's the asymmetry with SendFullServerSnapshot right above it in the same file.

The re-send — NetworkObject.cs:441 (UpdateTransmitState), the per-tick cull evaluation. For the new connection a CullState is created lazily (:479-486). The moment the object goes through one cull → visible-again cycle for that connection, it hits:

```cs
if ( !_culledConnections.Remove( target.Id ) )   // :498
    continue;
LocalSnapshotState.RemoveConnection( target.Id );
GameObject.Network.SetCullState( target, false );
EnsureCreateMessageSent( target );               // :503 ← fires because the set is empty
```
Since the join snapshot never marked the connection, EnsureCreateMessageSent believes this client has never seen the object and ships a second full create for a GUID the client already registered at join → GameObjectDirectory.Add collides on the GameObject and every component → your warning spam, one line each for waterramp01/ModelRenderer/ModelCollider/Rigidbody/Buoyancy, and a duplicated prop. (The same unguarded call exists at :452 for the AlwaysTransmit-toggled-on case.)

The cull/uncull cycle happens almost immediately for a joiner in your game because the second client sits in character select with no pawn — IsVisible (NetworkObject.cs:552) falls through to Scene.IsBBoxVisibleToConnection, whose answer for a pawnless connection can differ from what the snapshot decided, so the object gets culled during loading and unculled the moment the player spawns — hence the burst right at join.

The one-line fix: in GetSnapshot, after SerializeNetworkObjects(source, msg.NetworkObjects), walk msg.NetworkObjects and call MarkCreateMessageSent(source) for each included object — literally the same loop that already exists in SendFullServerSnapshot at SceneNetworkSystem.cs:319-327. (A belt-and-braces alternative on the receive side: have the client treat an ObjectCreateMsg for an already-known GUID as a refresh instead of a new object.)

That's a tight, verifiable writeup for the issue report — quote the two paths and line numbers and it should be an easy fix on their end.

PR #11406, opened 2026-07-08
Lower-Level Microphone Access

kurozael:
# Pull Request

---

  ## Summary

  Replaces the internal `VoiceManager` with a new public `Microphone` class, giving games low-level
  access to voice capture: record Steam-compressed voice data via a callback, transmit it however
  they like, and decode it back into PCM samples on the receiving end.

  Previously voice capture was only usable through the plug-and-play `Voice` component. Games that
  wanted custom voice transport (own networking, voice recording, processing, etc.) had no way to
  get at the raw compressed data.

  ## Motivation & Context

  Games want full control over voice. I specifically would like it in my game VAST because it uses custom networking and I'd like to have full control over how I transmit the data. Capture, send through their own transport, decode on the
  other side. The `Voice` component bundles capture, networking and playback together, so there was
  no supported way to do any of these steps yourself. `Microphone` exposes the capture/decode layer
  that `Voice` (and menu party voice) are now built on top of.

This would also allow people to do any modifications they wanted to the actual voice data, if they wanted to apply effects to it for example.

  Fixes: N/A

  ## Implementation Details

  - **`Microphone` (public, `Sandbox.Engine/Systems/Audio/Microphone.cs`)** — renamed/reworked from
    the internal `VoiceManager`. Public surface: `StartRecording()` / `StopRecording()`,
    `IsAvailable` / `IsRecording` / `IsActive`, `SampleRate`, `OnCompressedData`, and
    `Decompress()` for turning received compressed data back into 16-bit PCM.

  - **`OnCompressedData` is a settable callback, not an event.** This keeps cleanup simple and
    safe: when the game closes, `GameInstanceDll.ResetEnvironment()` calls `Microphone.Reset()`,
    which just nulls the callback so an unloaded game's handlers can't keep its assemblies alive
    or keep receiving mic data after it's gone. It also stops recording if nothing is listening
    anymore, so a game can't leave the mic hot on exit.

  - **Engine systems use a separate internal event** (`OnCompressedDataInternal`). `PartyRoom` and
    `VoiceComponent` subscribe to that instead of the public callback, so menu party voice survives
    game unload while game callbacks are wiped.

  - `SoundStream.WriteData` now takes `ReadOnlySpan<short>` instead of `Span<short>` so decoded
    (read-only) sample views can be written directly.

  - `Application.IsMicrophoneListening/IsMicrophoneRecording` map to the clearer
    `Microphone.IsRecording/IsActive` names internally; behaviour unchanged.

  - Callback docs note that the supplied memory is a view into a reused internal buffer and must be
    copied if kept beyond the callback (same contract the old `VoiceManager` had implicitly).

  N/A — no UI/rendering changes.

  ## Checklist

  - [ ] Unit tests added where applicable and all passing — none added; voice capture needs Steam
        running, so it isn't practical to unit test


boxrocket6803:
regarding previous concerns facepunch had about recording it might be a good idea to move the logic for activating the mic icon into this class so we still have control but bad actors wouldn't be able to do anything more than they already could (with the existing system one could write an audio processor to record then silence audio in the voice channel, the mic icon is the main protection against that)

kurozael:
> regarding previous concerns facepunch had about recording it might be a good idea to move the logic for activating the mic icon into this class so we still have control but bad actors wouldn't be able to do anything more than they already could (with the existing system one could write an audio processor to record then silence audio in the voice channel, the mic icon is the main protection against that)

I *think* this solves that?

PR #11416, opened 2026-07-09
Physics: sync only bodies the solver reports as moved

danfmn:
## Summary

`ScenePhysicsSystem.UpdatePhysics` scanned every `Rigidbody` each physics step and called `UpdateTransformFromBody()` on all of them, doing a native transform read per body even for the many that are asleep/settled. That's O(all bodies) on the main thread every step, regardless of how many actually moved (there was a standing "todo: skip asleep?" here).

This change syncs only the bodies the solver reports as moved. Cost now scales with *moved* bodies instead of *total* bodies; asleep bodies cost nothing.

| Bodies (all asleep) | before | after | reduction |
|---------------------|--------|-------|-----------|
| 6000 | ~3.5 ms/step | ~0.8–1.3 ms/step | **~65–77%** |

*(Measured via `PerformanceStats.Timings.Physics`, median of per-step samples, all bodies settled. Master was consistent at ~3.5 ms; the branch varied 0.8–1.3 ms across runs.)*

## Motivation & Context

Scenes with many sleeping bodies such as settled debris, prop piles, destruction gibs paid a per-step transform-sync cost proportional to the *total* body count even when nothing was moving. The loop was pure waste for asleep bodies, and it runs on the main thread every step.

Fixes: 

## Implementation Details

The Native solver already fires a per-body callback only for bodies that moved this step (`PhysicsEngine.OnActive` → `PhysicsBody.OnActive`). This PR surfaces that as a new internal `PhysicsWorld.OnBodyActive` event and drives the sync from it:

- **`PhysicsWorld`** — new internal `Action<PhysicsBody> OnBodyActive`.
- **`PhysicsBody.OnActive`** — after `Dirty()`, invoke `World?.OnBodyActive?.Invoke( this )`.
- **`ScenePhysicsSystem`** — collect moved bodies into a `HashSet`, then drain only that set after the step instead of scanning `Scene.GetAll<Rigidbody>()`.

**Before (master):**

https://github.com/user-attachments/assets/4999a590-77c8-4821-87ae-a091250ba254

**After (this branch):**

https://github.com/user-attachments/assets/7af5c3f2-9d38-4f86-92d0-49fad28e9301

## Checklist

### DIFF
--- a/engine/Sandbox.Engine/Scene/GameObjectSystems/ScenePhysicsSystem.cs
+++ b/engine/Sandbox.Engine/Scene/GameObjectSystems/ScenePhysicsSystem.cs
@@ -13,6 +13,11 @@ sealed class ScenePhysicsSystem : GameObjectSystem<ScenePhysicsSystem>
 	private HashSet<Rigidbody> RigidBodies { get; set; } = new();
 	private List<ISceneCollisionEvents> CollisionEvents { get; } = new();
 
+	// Rigidbodies whose physics body moved this step (populated from PhysicsWorld.OnBodyActive).
+	// We drain this each step instead of scanning every Rigidbody, so settled/asleep bodies cost nothing.
+	private readonly HashSet<Rigidbody> MovedRigidbodies = new();
+	private readonly List<Rigidbody> MovedBuffer = new();
+
 	internal bool Enabled { get; set; }
 
 	public ScenePhysicsSystem( Scene scene ) : base( scene )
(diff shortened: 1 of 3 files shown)


KivinDev:
Bump! I have hundreds of Rigidbodies in my scenes, and they're taking up a lot of frame time, because of this step.

lolleko:
Hello,

I think this makes sense, as the todo stated it's something we wanted to do.
However, we will likely do this optimization ourselves as this is something we have to properly benchmark.

In the meantime if this is urgent to you you can build the engine from source.

lolleko:
I cannot reproduce your numbers, your changes make no measurable difference in our benchmarks.
What are your system specs, how is you test scene structured? is it the same as in the video? You should always run benchmarks outside the editor.

danfmn:
> I cannot reproduce your numbers, your changes make no measurable difference in our benchmarks. What are your system specs, how is you test scene structured? is it the same as in the video? You should always run benchmarks outside the editor.

Hi lolleko, thanks for taking a look. 

Specs:
NVIDIA GeForce RTX 4070 Ti SUPER
AMD Ryzen 7 7800X3D
64 GB DDR5 RAM

The test scene is the one shown in the video. I did not run it outside the editor unfortunately. You can view the script I used to setup the scene here: https://gist.github.com/danfmn/52b460c674870523f00a95b3ab31ed62

The delta should only be visible when a large body count is asleep. While anything is moving, both paths do the same work. There might be better ways to validate this given that.

PR #11420, opened 2026-07-09
Nest asset creation categories

TheDimin:
# Pull Request

---

## Summary

Makes the Asset Browser `New` menu support slash-delimited asset categories as nested folders.

For example, `Category = "Survival/Crafting"` now appears as:

`New -> Survival -> Crafting -> Crafting Recipe`

instead of a flat `Survival/Crafting` submenu.

## Motivation & Context

Custom asset types can use category strings to organize creation options, but slash-delimited categories were previously shown as one flat menu label. This makes larger projects harder to scan when they group resource types into nested domains.

## Implementation Details
- Normalizes category strings by splitting on `/`, trimming empty segments, and joining them back before sorting/grouping.
- Walks each normalized category segment through `FindOrCreateMenu` to create nested folders.
- Sorts category groups so nested folders are created before loose items at the same level, keeping folders above direct options.
- Ignores null, empty, and whitespace-only categories in the categorized section.

After:
image
(Bow weapon is placed in survival folder for ilustration that paths without `/` still work )
Before:
image

## Checklist

### DIFF
--- a/game/addons/tools/Code/CreateAsset.cs
+++ b/game/addons/tools/Code/CreateAsset.cs
@@ -96,10 +96,16 @@ public static class CreateAsset
 
 		parent.AddSeparator();
 
-		var grouped = entries.OrderBy( x => x.Name ).GroupBy( x => x.Category ).OrderBy( x => x.Key );
+		var grouped = entries.OrderBy( x => x.Name )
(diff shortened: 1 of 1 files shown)


aylaylay:
We have this now

TheDimin:
https://github.com/Facepunch/sbox-public/commit/4843aa7f7cd022e6e779d0773651f9f07be31e42

PR #11428, opened 2026-07-09
Add performance overlay and perf_dump command

1igaming:
Two small tools for measuring performance without a profiler attached:

- **`perf_overlay` (saved menu convar)** — a compact always-on-top overlay in the corner showing FPS, average/worst frame time, GPU time, draw calls and VRAM used/budget. Rendered from the existing `PerformanceStats`/`FrameStats` counters, reuses the DevUI `StatValue` component, mounted in `DevLayer` next to the console overlay. Costs nothing when hidden (BuildHash returns 0).
- **`perf_dump [seconds] [quit]`** — samples frame time, GPU time, draw calls and triangles every frame for N seconds (default 10), then writes a JSON report to `logs/` with min/avg/max/p95/p99 summaries, FPS stats including 1% lows, VRAM usage, and a snapshot of the relevant graphics convars so reports are comparable. Takes a screenshot when it finishes, and optionally quits the game afterwards so it can be scripted from the command line for automated before/after runs.

Self-contained; no engine changes.

### DIFF
--- a/game/addons/menu/Code/DevUI/DevUI.cs
+++ b/game/addons/menu/Code/DevUI/DevUI.cs
@@ -8,6 +8,7 @@ public class DevLayer : RootPanel
 	{
 		AddChild<DeveloperMode>();
 		AddChild<ConsoleOverlay>();
+		AddChild<PerformanceOverlay>();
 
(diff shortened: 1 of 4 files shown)


lolleko:
We already have several overlays that show exactly this information.

PR #11429, opened 2026-07-09
Add render resolution slider, advanced quality settings and menu background modes

1igaming:
**Depends on #11428** — the "Performance Overlay" row in VideoSettings.razor binds to `Sandbox.UI.Dev.PerformanceOverlay.ShowPerfOverlay`, which is added there. That branch is now merged into this one so CI can build it standalone, but #11428 should still land first (the merge will deduplicate).

The video settings page currently exposes a handful of coarse quality groups. On low-end hardware (integrated GPUs in particular) that's not enough control — the difference between playable and not is usually one or two specific features, not a whole tier.

This adds:

- **Render Resolution slider (25–100%)** shown for all upscaler modes except FSR3 (which manages its own scale through quality presets). At 100% the upscaler switches off; below that FSR1 kicks in unless Stretch/FSR1 was picked explicitly. The engine-side clamp on `upscaler.render_scale` is relaxed from 0.4 to 0.25. Picking Off snaps the slider back to 100% so the two controls can't contradict each other.
- **Advanced section** with individual controls that were previously only reachable via console: shadow distance, sun shadows, dynamic light shadows, contact shadows, ambient occlusion quality, SSR quality, bloom, decal limit. Getters read the live convar so the UI always reflects reality; setters store a cookie so an explicit choice survives restarts and quality-group changes (`RenderSettings.ApplyAdvancedOverrides` re-applies cookies on top of the group fan-out). Settings the user never touched keep following their group.
- **Startup re-apply in Bootstrap**: `RenderSettings` can be constructed before the engine convars are registered, which silently drops its writes. After `InitEngineConVars()` we now re-run `SetDefaults` + `ApplyAdvancedOverrides` + `ApplyUpscalerSettings` so saved settings actually take effect on boot. The upscaler part matters: the native engine restores `machine_convars.vcfg` at startup, and without the re-apply the saved upscaler choice (including Off) never reached `r_upscaling` — the vcfg value silently won.
- **Menu Background setting (Full / Simple / Off)** — a new option controlling the home menu background. Simple turns off spotlight shadows, bloom, AO and painted clutter in the menu scene (snapshot/restore, nothing is lost when switching back to Full); Off disables the background scene entirely. Backed by the saved `menu_background` convar.
- **Low quality profile cuts** in `quality_profiles.json`: DoF and AO off on Low, and the shadow tiers now also set `r.shadows.csm.distance` and `r.shadows.max` (4000/16, 9000/64, 15000/256).

Measured on an Intel iGPU in the menu scene: the advanced knobs take GPU frame time from 33.5 ms to 23.8 ms without dropping the whole quality tier.

### DIFF
--- a/engine/Sandbox.Engine/Core/Bootstrap.cs
+++ b/engine/Sandbox.Engine/Core/Bootstrap.cs
@@ -176,9 +176,11 @@ internal static class Bootstrap
 			InitEngineConVars();
 
 			// RenderSettings may be constructed before these convars exist, silently dropping
-			// its writes - re-apply the quality groups and advanced overrides now
+			// its writes - re-apply the quality groups and advanced overrides now. The upscaler
+			// settings also need pushing here so the cookies win over machine_convars.vcfg.
 			Settings.RenderSettings.Instance.Config.SetDefaults( Settings.RenderSettings.Instance );
 			Settings.RenderSettings.Instance.ApplyAdvancedOverrides();
+			Settings.RenderSettings.Instance.ApplyUpscalerSettings();
 
 			if ( IToolsDll.Current is not null )
 			{
--- a/engine/Sandbox.Engine/Core/Bootstrap.cs
+++ b/engine/Sandbox.Engine/Core/Bootstrap.cs
@@ -175,6 +175,11 @@ internal static class Bootstrap
 
 			InitEngineConVars();
 
+			// RenderSettings may be constructed before these convars exist, silently dropping
+			// its writes - re-apply the quality groups and advanced overrides now
+			Settings.RenderSettings.Instance.Config.SetDefaults( Settings.RenderSettings.Instance );
+			Settings.RenderSettings.Instance.ApplyAdvancedOverrides();
+
 			if ( IToolsDll.Current is not null )
 			{
 				using var x = StartupTiming?.ScopeTimer( $"IToolsDll Bootstrap Init" );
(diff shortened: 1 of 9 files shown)


1igaming:
The tests check was failing because this branch didn't build without #11428 (NoWarnings caught the missing PerformanceOverlay reference, and MenuInitialization failed downstream of the same compile error). I've merged that branch into this one so CI builds it standalone.

While retesting I also found a real bug: the saved upscaler settings never reached their convars at startup. The native engine restores machine_convars.vcfg on boot, and the Bootstrap re-apply only covered the quality groups and advanced overrides - so if the vcfg and the cookies disagreed, the vcfg won and turning the upscaler off didn't stick across sessions. Fixed by pushing the upscaler cookies to their convars in the same startup re-apply, and picking Off in the UI now snaps the render resolution slider back to 100% so apply can't flip it back on. Verified with perf_dump before/after: cookie set to Off with vcfg saying FSR1 used to leave r_upscaling at 2, now it comes up as 0.

1igaming:
The latest tests run died during artifact download, before any tests ran - plump_lips_ao_png_8b304a3.generated.vtex_c came back with a hash mismatch from artifacts.sbox.game three times in a row. Looks like CDN trouble rather than anything in this diff. I can't re-run the job from a fork, so a re-run from someone with access would be appreciated.

1igaming:
Reopening re-triggered the tests run and it failed the same way - plump_lips_ao_png_8b304a3.generated.vtex_c hash mismatch (ca1ea037...) three attempts, before any test ran. Same failure just hit an unrelated PR too (physics-sleep-aware-transform-sync, run 29058419862, same file and hash), so the CDN copy of that artifact looks corrupt rather than flaky. Re-runs probably won't pass until that artifact is fixed on artifacts.sbox.game.

1igaming:
Found that FSR1 (EASU) is only specified up to 2x per axis - below 50% render scale it smears badly and RCAS turns the undersampling into speckle. Render scale is now floored at 50% for FSR1, and the slider falls back to Stretch for anything lower.

PR #11430, opened 2026-07-09
Add downloaded games list to the storage settings page

1igaming:
The storage settings page shows total cache usage by file type, but there's no way to see which games are taking the space or to free up space per game — your only option is nuking the whole cache.

- **`DownloadedPackages` (engine)** keeps a small JSON record per package in `download/.packages/` — ident, title, type, thumb, version, download/last-used times and the file manifest (path, CRC, size). `Package.DownloadAsync` refreshes the record on every download or cache mount, so records converge for packages downloaded before this change after they're played once.
- **`MenuUtility.Storage`** exposes the records to the menu: per-package current cache usage (only counts files actually on disk right now), per-package file lists for the type breakdown, and deletion.
- **Storage page** gets a "Downloaded Games" list — thumbnail, title/ident, per-type file breakdown, download/last-played dates, size on disk — sortable by size, last played or name. Non-game packages (clothing, models etc) are rolled up into a single "Avatar Items & Shared Assets" row.
- **Deleting is always safe:** it only removes cached files, never anything authoritative — the package re-downloads on demand next time it's played (that's how the cache already works). Files that another recorded package also references are left alone. Delete asks for confirmation and states the size it will free.

Self-contained; no dependency on the other PRs.

### DIFF
--- a/engine/Sandbox.Engine/Services/Packages/DownloadedPackages.cs
+++ b/engine/Sandbox.Engine/Services/Packages/DownloadedPackages.cs
@@ -1,0 +1,143 @@
+using System.IO;
+
+namespace Sandbox;
+
+/// <summary>
+/// Keeps a record of every package we've downloaded into the asset cache, so we
+/// can attribute cached files to packages later (storage usage, cleanup etc).
+/// Records live alongside the cache in download/.packages/.
+/// </summary>
+internal static class DownloadedPackages
+{
+	const string Folder = ".packages";
+
+	internal class Record
+	{
+		public string Ident { get; set; }
+		public string Title { get; set; }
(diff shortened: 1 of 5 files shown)


lolleko:
Can you post a screenshot of how this actually looks like.

1igaming:
Ok

1igaming:
image

1igaming:
Refined Storage dates (labeled **Downloaded** / **Last played** rows) and footer button sizing, plus a session cache so reopening Storage is fast — stale scans refresh quietly in the background; Refresh / delete / Delete Unused still force a rescan.

Storage settings

PR #11436, opened 2026-07-10
Adding grid to orthographic editor views

### LINKED ISSUE #11405: Navigation Gizmo Grid
### What can't you do?

When you use the Navigation Gozmo in the top right of the viewport, there is no grid on the view that you should at that flat orentiation. There should be a grid, like in blender.

When you select one of the flat/2d views using the navigation gizmo, there should be a grid (like the one that by default is on the floor) on that orientation.

Another possible feature for this would be the option to make it draw behind or infront of map geomtry, etc.

### What have you tried?

Using the Gizmo

### Additional context

Would just be a super helpful feature.

A-PS1999:
# Pull Request

---

## Summary

Unlike other game engines and 3D modeling tools like Maya or Blender, right now in S&box's editor, switching to an orthographic view along one of the axes (X, Y, Z) doesn't make the grid appear in that view. Currently the grid only exists on the XY plane in the regular perspective view.

This change makes the S&box editor behave in line with other industry tooling. The grid being visible in the orthographic views can make work easier (placing objects appropriately, etc).

## Motivation & Context

Largely as above, but I was motivated to do this by this issue which was raised:
https://github.com/Facepunch/sbox-public/issues/11405

Fixes: #11405 

## Implementation Details

In `SceneViewportWidget.OrientationGizmo.cs` within `SnapToAxis` I added code that checks which is the axis that the user has selected and uses a switch statement to select the appropriate grid axis to use. I subsequently set `State.GridAxis` to this.

When the user drags on the orientation gizmo to leave the orthographic view, in `OrbitAroundPivot`, I set `State.GridAxis` back to `Gizmo.GridAxis.XY`.

## Checklist

### DIFF
--- a/game/addons/tools/Code/Scene/SceneView/SceneViewportWidget.OrientationGizmo.cs
+++ b/game/addons/tools/Code/Scene/SceneView/SceneViewportWidget.OrientationGizmo.cs
@@ -1,4 +1,4 @@
-namespace Editor;
+﻿namespace Editor;
 
 public partial class SceneViewportWidget
 {
@@ -146,12 +146,7 @@ public partial class SceneViewportWidget
 
 	private void OrbitAroundPivot()
 	{
-		_gizmoOrthoActive = false;
-
-		if ( State.Is2D )
-			State.View = ViewMode.Perspective;
-
-		cameraTargetPosition = null;
+		EnterPerspectiveView();
 
 		var delta = Application.CursorDelta * 0.1f;
 		GizmoInstance.OrbitCameraAroundPivot( _activeCamera, delta, ref cameraOrbitDistance );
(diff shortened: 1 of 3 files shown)


CarsonKompon:
This looks good, but should it really be in `OrbitAroundPivot()`?

A-PS1999:
> This looks good, but should it really be in `OrbitAroundPivot()`?

Thanks. I guess I was trying to make it the smallest fix I could that touched the least amount of other code but you're right, the logic ought to be better consolidated in its own function(s). I see that you've self-assigned but do you mind if I make an attempt at that change? If it's still not satisfactory after that then please by all means make changes as you feel appropriate.

PR #11451, opened 2026-07-11
Adds ForceVisibilityUpdate which bypasses cull-delay grace period on hiding

### LINKED ISSUE #10982: Network Visibility doesn't update instantly when an object is no longer visible
### Version / Commit

26.05.13a

### Describe the bug

`INetworkVisible.IsVisibleToConnection` doesn't replicate instantly to other connections.

e.g. If you want to hide the player who just died from other connections, it will take about 800 milliseconds or so to take effect on the network. Clearly some latency issues there.

### To Reproduce

Here's an example class I have:

```cs
    [Icon("disabled_visible")]
    public class PlayerVisibility : Component, Component.INetworkVisible, IPlayerSystem {
        public PlayerData Data { get; set; }
        
        public bool IsVisibleToConnection(Connection conn, in BBox worldBounds) {
            if (!PlayerManager.Instance.Players.TryGetValue(conn.Id, out var other)) return false;
            return other.Health.Dead || !Data.Health.Dead;
        }
    }
```
This should hide dead players for alive players instantly, but an alive player can see the dead player for a split second.

(I'm not sure if it's because sync variables are taking a while cause Dead is a sync variable, or the visibility system is just slow)

### Expected behavior

Network visibility should take effect as fast as possible.

### Media/Files

_No response_

### Additional context

_No response_

deadmonstor:
# Pull Request

---

## Summary

Adds a ForceVisibilityUpdate() method that lets you force an immediate re-evaluation of network visibility for an object, instead of waiting out the normal cull-delay grace period.

## Motivation & Context

Fixes: #10982 

## Implementation Details

Intentionally opt-in rather than automatic because I think in general people are going to want the culling delay on more than off. So this is a function to shortcut that.

Both Before and After: https://www.youtube.com/watch?v=mphzAjutAmY
Code: 
```csharp
using Sandbox;

public sealed class Cube : Component, Component.INetworkVisible
{
	[ConVar] public static bool cube_enable { get; set; } = true;
	[ConVar] public static bool cube_enable_with_new_function { get; set; } = false;

	private bool _isVisibleToConnection = true;
	public bool IsVisibleToConnection( Connection conn, in BBox worldBounds )
	{
		if ( _isVisibleToConnection != cube_enable && cube_enable_with_new_function )
		{
			GameObject.Network.ForceVisibilityUpdate();
		}

		_isVisibleToConnection = cube_enable;
		return cube_enable;
	}
}
```

## Checklist

### DIFF
--- a/engine/Sandbox.Engine/Scene/GameObject/GameObject.Network.cs
+++ b/engine/Sandbox.Engine/Scene/GameObject/GameObject.Network.cs
@@ -790,6 +790,23 @@ public partial class GameObject
 		}
 
 		/// <summary>
+		/// Force an immediate re-check of network visibility for this object, bypassing the normal cull-delay
+		/// grace period. Call this when your <see cref="Component.INetworkVisible"/> implementation's result
+		/// changes and you want the new visibility state (e.g. hiding a dead player) to take effect on the very
+		/// next network tick instead of after the usual grace period.
+		/// <br/>
+		/// <br/>
+		/// Note: Do not call this every tick, only call it when you need the visibility state to be re-evaluated immediately.
+		/// </summary>
+		public void ForceVisibilityUpdate()
+		{
+			if ( !Active || (IsProxy && !Networking.IsHost) )
+				return;
+
+			go._net?.ForceVisibilityUpdate();
+		}
+
+		/// <summary>
 		/// Become the network owner of this object.
 		/// <br/>
 		/// <br/>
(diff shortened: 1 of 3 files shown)


lolleko:
I do wonder what purpose the delay has. Maybe we can get rid of it instead?

deadmonstor:
> I do wonder what purpose the delay has. Maybe we can get rid of it instead?

Yeah I guess the only thing that would be annoying is if we true/false spam it will end up sending crazy amounts of data over the network, maybe its just better to allow turning off on certain network objects instead of a code solution

lolleko:
Maybe it could be turned off in general? if some true/false spams it it's really their issue, we also don't prevent people from spam recreating gameobjects & components.

My only concern would be that this could be abused by a malicious actors, but there are probably plenty other ways to DoS a server.

deadmonstor:
Honestly I think we could just remove the whole system and the player could implement it but its probably better for a disable project wide honestly

lolleko:
What are you referring to by whole system?

PR #11456, opened 2026-07-11
Fix box & hull collider center offset ignoring collider rotation rela…

solrmax:
# Pull Request

---

Fixes `BoxCollider` and `HullCollider` producing a physics shape offset from their gizmo when the collider is rotated relative to its parent `Rigidbody`.

The `Center` offset was translated into the body's space but never rotated, causing the actual `PhysicsShape` to sit in the wrong spot while the green editor gizmo (drawn in the collider's own space) stayed correct: visible as **two boxes** when `Center != (0,0,0)` and the a rigidbody is involved.

## Motivation & Context

Any rotated child GameObject with a `BoxCollider` or `Hull` box under a `Rigidbody` parent has its collision volume silently misplaced relative to what's shown in the editor, since collision math never applied `local.Rotation` to `Center`.

Disabling the parent `Rigidbody` (which makes the local transform identity) hide the bug.

## Implementation Details

`SphereCollider` and `CapsuleCollider` already transform their center via `local.PointToWorld(...)`, which correctly rotates the local offset into the `Rigidbody`'s space before translating it. While `BoxCollider` and `HullCollider` only doing:

```csharp
box.Mins += local.Position;
box.Maxs += local.Position;
```

a plain translation with no rotation applied to the offset which is correct only when the collider's rotation matches the Rigidbody's. Changed both `UpdateShape` and `CreatePhysicsShapes` in `BoxCollider.cs` and `HullCollider.cs` to use `local.PointToWorld( Center )` instead, matching the existing idiom (in `SphereCollider`).

BEFORE (see the box collider move as it rotates while the sphere collider remains completely still)
https://github.com/user-attachments/assets/0dc5832e-ee62-4f9d-bf61-89f2a82232d9

AFTER (every colliders are completly still while rotating)
https://github.com/user-attachments/assets/0640b6fd-4167-4b89-ad12-038251a920bb

## Checklist

- [ ] Public APIs are documented (if applicable) — n/a, no public API changed

### DIFF
--- a/engine/Sandbox.Engine/Scene/Components/Collider/BoxCollider.cs
+++ b/engine/Sandbox.Engine/Scene/Components/Collider/BoxCollider.cs
@@ -81,13 +81,9 @@ public sealed class BoxCollider : Collider
 		var body = Rigidbody;
 		var world = Transform.TargetWorld;
 		var local = body.IsValid() ? body.Transform.TargetWorld.WithScale( 1.0f ).ToLocal( world ) : global::Transform.Zero;
-		var box = BBox.FromPositionAndSize( Center, Scale );
-		box.Mins *= world.Scale;
-		box.Maxs *= world.Scale;
-		box.Mins += local.Position;
-		box.Maxs += local.Position;
+		var center = local.PointToWorld( Center );
 
-		Shape.UpdateBoxShape( box.Center, local.Rotation, box.Size * 0.5f );
+		Shape.UpdateBoxShape( center, local.Rotation, Scale * world.Scale * 0.5f );
 
 		CalculateLocalBounds();
 	}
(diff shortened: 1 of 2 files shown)


solrmax:
image

In this screenshot, you can see a box (parent) and a box collider (a game object that is a child of the big box). 
The parent has a Rigidbody, and Show Physics Debug is enabled. 
If you select the game object containing the box collider, you can see that the gizmo and the actual collider are not aligned (green vs gray boxes).

image
image
image

lolleko:
We need to check if this breaks existing behaviour.

solrmax:
> We need to check if this breaks existing behaviour.

Yeah of course, and maybe my changes aren't the most elegant.

PR #11467, opened 2026-07-12
POC: Docked in-process multiplayer clients in the editor

Ardivaba:
**This is a proof of concept pull request, not meant to be merged, but exists for people running a custom engine build and want to use this feature.**

### Demonstration
https://www.youtube.com/watch?v=iVsJQ4yXxi8

### Description

Adds **docked in-process multiplayer clients** to the editor: "Client 1..N" tabs, each a genuine multiplayer client living inside the editor process, connected to the in-editor host through an in-memory transport and the real connection handshake. Joining is near-instant (no process spawn, no code archive recompile, no file downloads), making it practical to iterate on multiplayer with several clients open — or stress a host with many.

### How it works

- **In-memory transport** (`InProcessSocket`/`InProcessConnection`): the host gets a socket like any other, clients go through ServerInfo → UserInfo → Welcome → Snapshot → Activate, so game code sees a real client with a distinct (fake) identity. Handshake steps that would tear down shared process state (package load, code archives, mounts) are skipped via `NetworkSystem.IsInProcessClient`.
- **Per-client world isolation** (`InProcessTenant`): each session owns a `GlobalContext` (the same seam that separates Menu and Game today) plus **private copies of the game assemblies in a collectible `AssemblyLoadContext`**, loaded from the host's already-compiled bytes. Game statics are per-client — a game that keeps state in statics behaves like it would across real processes. Tenant TypeLibrary/NodeLibrary/ResourceSystem/UISystem/Json options are per-world; engine assemblies stay shared by identity. Falls back to shared mode on any failure, toggleable with `docked_client_isolation`.
- **Hotload**: when host code hotloads, sessions rebuild — an instant in-process reconnect with the new assemblies.
- **Teardown is surgical and leak-free**: tenant assemblies fully unload (verified by tests) via explicit `LoadContext` child unload, per-assembly pruning of `ReflectionQueryCache`/`MemberCopyCache`/`ReflectionCache`, and per-instance Json cache clearing.
- **Input**: the focused docked client is a first-class participant in `InputRouter` — its own `InputContext` decides UI-vs-game routing and mouse capture, `Screen` and `Mouse.Visibility` are per-world, and focus transitions hand the engine window state between views in strict order. Green frame marks the focused tab; `Ctrl+F1..F12` switches focus between clients and back to the host; `Esc` releases.
- **Audio**: each client gets a submix under Master whose volume follows tab focus; `Sound.StopAll` is scoped to the calling world.

### Pre-existing engine bugs fixed along the way

- `StringToken` reverse lookups returned whichever casing registered first (the hash is case-insensitive) — a type name like `Trigger` registering before the collision tag `trigger` broke case-sensitive tag comparisons. Reverse values are now canonical lowercase.
- `ResourceSystem.Unregister` removed entries by id without checking the instance — destroying any second wrapper of a file evicted the original registration. Now instance-checked.
- `SceneNetworkSystem.Dispose` leaked `BatchSpawnList`/`PendingSceneLoads` references into destroyed scenes.
- Stylesheet file watchers could fire against torn-down contexts (NRE).

### Tests

`Sandbox.Test.Integration/Scene/Network/InProcessIsolation.cs` covers the full handshake, multi-client identity, static-field isolation across tenants (a probe assembly compiled at test time, loaded once per world), tenant-typed components arriving through the real snapshot, camera survival, replication to isolated clients, rebuild-on-hotload flagging, host-resource survival across tenant disposal, and full ALC unload (connected and not).

### Known limitations

- Clicking UI inside an unfocused client tab does nothing (by design — one input processor per frame).
- Tenant game resources don't hot-reload on asset file changes; sessions rebuild on code hotload.
- Real Steam auth, crash isolation and honest per-client performance still require out-of-process instances (`-joinlocal` unchanged).


lolleko (changes requested):
Can you find a way to make this less verbose and figure out another way than
`if ( !GlobalContext.Current.IsInProcessTenant )` everywhere?

Like this idea but it should be self contained and ideally not leak too much into game code.

Ardivaba:
@lolleko Done.

antopilo:
@Ardivaba This seems way too intrusive of a change

Ardivaba:
@antopilo Agreed, but I don't see a way to make it less intrusive (might be a skill issue).

Question is whether this particular PR has any mergeable future at all, or if it's best left as a usable proof of concept pr for people willing to build custom engine version.

It is usable and it does provide a substantial productivity boost, so the path might be to maintain it until proper version of this gets implemented.

PR #11479, opened 2026-07-13
Flags preserved at runtime on prefab instances

### LINKED ISSUE #11425: Static flag not preserved at runtime on prefab instances
### Version / Commit

Latest public build as of 2026-07-09

### Describe the bug

I have a prefab marked as Static in the editor. When I place an instance of that prefab in a scene, it still appears as static in the editor.

However, at runtime, the prefab instance is no longer marked as static.

Regular scene GameObjects seem to keep the Static flag correctly, so the issue appears to happen specifically with prefab instances.

### To Reproduce

1. Create or open a prefab.
2. Mark the prefab as Static in the editor.
3. Place an instance of that prefab in a scene.
4. Verify that it appears as static in the editor.
5. Run the game.
6. Check the Static flag at runtime.

### Expected behavior

The prefab instance should remain marked as static at runtime.

### Media/Files

Image

Image

### Additional context

This seems to happen only with prefab instances, not with regular GameObjects created directly in the scene.

deadmonstor:
# Pull Request

---

## Summary

Persisted flags like static, that were set on a prefab root were dropped on prefab instances loaded from disk. This re-reads them so instances keep them at runtime.

## Motivation & Context

Fixes: #11425

## Implementation Details

I double checked that flags are the only thing that get lost because its the only thing above, just opted to reapply them instead of shuffling around 30 callsites.

Before: https://youtu.be/qCTVTaP2sEs
After: https://youtu.be/CLl-3CntYqI

## Checklist

### DIFF
--- a/engine/Sandbox.Engine/Scene/GameObject/GameObject.Serialize.cs
+++ b/engine/Sandbox.Engine/Scene/GameObject/GameObject.Serialize.cs
@@ -325,9 +325,6 @@ public partial class GameObject
 			JsonUpgrader.Upgrade( serializedVersion, node, GetType() );
 		}
 
-		DeserializeFlags( node, options );
-		Flags |= GameObjectFlags.Deserializing;
-
 		if ( node[JsonKeys.EditorSkipPrefabBreakOnRefresh] is null )
 		{
 			_prefabInstanceData = null;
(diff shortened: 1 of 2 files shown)


rzkid:
this also fixes https://github.com/Facepunch/sbox-public/issues/901

lolleko (changes requested):
can we just move the existing DeserializeFlags call down below the patch instead of calling it twice? then it reads the final node like Name/Transform already do.

deadmonstor:
> can we just move the existing DeserializeFlags call down below the patch instead of calling it twice? then it reads the final node like Name/Transform already do.

Only reason I didnt do that to start with was because there are a few early outs that I dont know the consequences of moving the flags down but I can double check for you

lolleko:
Should be fine, if it early outs because the prefab is missing there isnt anything we can do anyway.

deadmonstor:
Ran all the tests to do with this and just moving it just below here works fine, should be good!

lolleko (approved):

PR #11491, opened 2026-07-15
Fix Scene map packages not loading clutter / mesh components

LtBrandon:
# Pull Request

## Summary

Allows BlobDataSerializer to locate binary blobs saved in package filesystems, and syncs the MapInstance scene loading path with Scene.Load. Without these changes your map won't load important data like clutter, baked nav meshes, and mesh component models.

## Motivation & Context

I wanted to publish a map using a Scene. When I tested it outside of the editor I found that none of my grass or mesh components were visible. I attempted a fix on the game side, but all of the data I would need to handle it nicely was marked as internal.

Fixes: N/A

## Implementation Details

Three notes:
1. I'm not 100% sure if I found the best method for getting the package's filesystem, it seems a little convoluted to me. I didn't immediately see a better way, but let me know if there is one.
2. I opted to leave out applying the scene properties when loading the MapInstance. When I looked at the list of properties that would be replaced it didn't seem desirable to update them. It's a quick change if you prefer I add it, though.
3. I added a warning when the data isn't found. This was a silent failure that took me a while to figure out. With the warning I would have found this in 5 minutes instead of 5 hours 😃 
 
After Changes:
image
image

Before Changes:
image
image

## Checklist

- [ x ] Code follows existing style and conventions
- [ x ] No unnecessary formatting or unrelated changes
- [ x ] Public APIs are documented (if applicable)
- [ x ] I’m okay with this PR being rejected or requested to change 🙂

### DIFF
--- a/engine/Sandbox.Engine/Scene/Components/Map/MapInstance.cs
+++ b/engine/Sandbox.Engine/Scene/Components/Map/MapInstance.cs
@@ -454,13 +454,13 @@ public partial class MapInstance : Component, Component.ExecuteInEditor
 		}
 
 		var props = sceneFile.SceneProperties;
-		
+
 		// Apply game object system properties
 		if ( props.TryGetPropertyValue( "GameObjectSystems", out var systemOverridesNode ) )
 		{
 			Scene.ApplyGameObjectSystemOverrides( systemOverridesNode );
 		}
-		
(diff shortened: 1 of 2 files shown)


PolEpie:
Similar to #11338

PR #11492, opened 2026-07-15
fix clutter grid system being executed in headless mode

solrmax:
# Pull Request

---

## Summary

Prevent `ClutterGridSystem` from registering its update and scene-load callbacks when s&box runs headless.

## Motivation & Context

Dedicated servers load scene systems even though they have no graphics output. `ClutterGridSystem` previously registered its callbacks in that environment, which could result in disposed `GpuBuffer<float>` errors during clutter updates.

Fixes: Can't find any issue related to that in gh.

## Implementation Details

Just add a guard `if (Application.IsHeadless) return;` in the constructor so the system never subscribes to `FinishUpdate` or `SceneLoaded`; this avoids unnecessary per-frame execution.

Not applicable. headless/dedicated-server-only change.

## Checklist


solrmax:
thats not the solution

PR #11455, opened 2026-07-11
Fix networked MeshComponent invisible on late-joining clients

deadmonstor:
# Pull Request

Thanks for contributing to **s&box** ❤️  
Please fill out the sections below to help us review your change efficiently.

---

## Summary

PolygonMesh geometry is stored as blob data, and reading it back needs the matching BlobDataSerializer scope open. In SetSnapshotAsync, objects deserialize inside a CallbackBatch, which defers property application until the batch flushes but each object's blob scope had already closed, so the mesh came back empty. ModelRenderer was unaffected (references a resource by path, no blob data).

## Motivation & Context

Fixes: #11453

## Screenshots / Videos (if applicable)

Before: https://youtu.be/rQXmdBViqLc
After: https://youtu.be/kYVfbwVzu3o

## Checklist

- [X] Code follows existing style and conventions
- [X] No unnecessary formatting or unrelated changes
- [X] Public APIs are documented (if applicable)
- [X] Unit tests added where applicable and all passing
- [X] I’m okay with this PR being rejected or requested to change 🙂

PR #11506, opened 2026-07-17
Fix MeshComponents loading empty from scene maps via MapInstance

ErazorWhite:
# Pull Request

## Summary

`MapInstance.LoadMapSceneGameObjects` deserializes a scene map's GameObjects without an open `BlobDataSerializer` scope. Every `{"$blob": guid}` reference resolves to `null` (`ReadBlob` returns null when no context is active), `PolygonMesh.JsonRead` quietly skips the null blob, and every `MeshComponent` instantiates with an **empty mesh** — silently: hierarchy, names and transforms are all correct, `LocalBounds` is zero, `MapInstance.IsLoaded` is true, nothing is logged. Non-blob components from the same scene (SkyBox2D, lights, ModelRenderer) load fine, which makes the geometry loss look like a content problem rather than an engine one.

Since PolygonMesh moved its geometry to blob data (proposed in #11324; on master in a reworked form), this hits **every scene-based map with MeshComponent geometry** loaded through `MapInstance` — including cloud map packages: a map republished with a current editor saves `$blob` refs and loads invisible in any game that mounts it.

The fix mirrors `Scene.Load` (`Scene.LoadSave.cs`), which opens the scope over the same deferred deserialize:

```csharp
using var sceneScope = Push();
using var blobs = BlobDataSerializer.Load( sceneFile.BinaryData, sceneFile.ResourcePath );
using var batchGroup = CallbackBatch.Batch();
```

Same invariant as #11455: the scope is declared before the `CallbackBatch` so it disposes after the batch flush, where the deferred property application actually reads the blobs.

## Motivation & Context

Hit in production with a cloud scene map (`rebornteam.zombie_city`, 332 MeshComponents): after it was republished with a post-blob-format editor, the map loaded as named empty objects for every player. Two things isolate the failure to blob-ref resolution rather than the data or the package:

- an older revision of the same package (saved before the blob format, geometry inline as JSON arrays) still renders through the unchanged legacy read path;
- rewriting the current scene's `"Data"` values from `$blob` refs to the inline base64 form renders correctly through this exact unpatched code path.

The compiled scene does ship the blob container (DBLOB block — verified present in the CDN-served `.scene_c`), so the `ResourcePath` fallback in `BlobDataSerializer.Load` reaches it for mounted packages, and `BinaryData` covers the in-memory case.

Repro:
1. Make a scene with MeshComponent geometry (editor mesh tools) and save it with a current build (`$blob` refs + `_d` sidecar).
2. Publish it as a map package.
3. In another project, point `MapInstance.MapName` at it.
4. The hierarchy loads; every mesh is empty (`LocalBounds` = 0,0,0), no errors logged.

Note: I couldn't build the engine locally to runtime-test the patch — the diagnosis and the fix pattern are validated as described above.

## Screenshots

map

## Checklist

- [x] Public APIs are documented (no API change)

### DIFF
--- a/engine/Sandbox.Engine/Scene/Components/Map/MapInstance.cs
+++ b/engine/Sandbox.Engine/Scene/Components/Map/MapInstance.cs
@@ -425,6 +425,11 @@ public partial class MapInstance : Component, Component.ExecuteInEditor
 
 		using var optionsScope = ActionGraph.PushSerializationOptions( sceneFile.SerializationOptions with { ForceUpdateCached = Scene.IsEditor } );
 		using var sceneScope = Scene.Push();
+
+		// Blob data (eg. MeshComponent geometry) is referenced by guid from the scene json -
+		// deserializing without an open scope silently yields empty blobs. Declared before the
+		// batch so it outlives the deferred property flush (usings dispose in reverse order).
+		using var blobs = BlobDataSerializer.Load( sceneFile.BinaryData, sceneFile.ResourcePath );
 		using var batchGroup = CallbackBatch.Batch();
 
 		foreach ( var json in sceneFile.GameObjects )


ErazorWhite:
Overlap note: #11338 already contains this exact MapInstance change as part of its clutter-publish fix — it predates the PolygonMesh blob migration, so its description doesn't mention meshes; since then the same missing scope also empties every MeshComponent of every scene map, which is what this PR isolates.

One gap this one-liner has for *mounted map packages*: `sceneFile.BinaryData` is already null here (`GameResource.LoadFromJson` consumes it and sets `BinaryData = null`), so resolution falls to `BlobDataSerializer.LoadFrom`, which only checks `FileSystem.Mounted` — and per #11338 that can't see package-mounted files ("else the _d wouldnt be loaded"). Its `ResolveBlobData` rework with the `PackageManager.MountedFileSystem` lookup covers that; local scene maps and in-memory SceneFiles work with this PR alone.

Happy to close this in favor of #11338, or extend it — whichever route you prefer. Flagging mainly that the mesh regression makes this path worth prioritizing.

ErazorWhite:
Closing in favor of #11338 — it contains this same MapInstance change plus the `ResolveBlobData`/`PackageManager.MountedFileSystem` lookup that mounted map packages additionally need. Leaving the diagnosis and repro above for reference.

PR #11524, opened 2026-07-18
Fix initialization order for editor clients connecting to dedicated servers

dictateurfou:
# Pull Request

## Summary

Fixes an initialization order regression when connecting to a **dedicated server from the editor**.

After the changes introduced in commit `8ca54b5dd03b96811ff13db659310ceb55ec5d36`, the editor client could enter the playing state before scene initialization had completed. As a result, RPCs sent from `OnActive(Connection channel)` could be received and executed before the client's `GameInstance.OnStart()` and `GameObject.OnStart()` callbacks had run.

This change delays entering the playing state until the client has fully initialized its scene.

## Motivation & Context

The regression was introduced by the changes in commit `8ca54b5dd03b96811ff13db659310ceb55ec5d36`, affecting the **editor → dedicated server** connection flow.

When user code sends an RPC from:

```csharp
public void OnActive( Connection channel )
{
    // Send RPC
}
```

the RPC can arrive on the editor client before scene initialization has completed, causing user code to execute before `OnStart()` has been called on the `GameInstance` and other `GameObject`s.

This results in an inconsistent initialization order that can lead to missing references or other startup issues.

Fixes: N/A

## Implementation Details

The editor client now enters the playing state only after scene loading and initialization have completed when connecting to a dedicated server.

This guarantees that:

* `GameInstance.OnStart()` has already executed.
* All `GameObject.OnStart()` callbacks have completed.
* RPCs sent from `OnActive(Connection channel)` cannot execute before the client is fully initialized.

The change only affects the timing of entering the playing state for the **editor → dedicated server** connection path.

N/A

## Checklist

### DIFF
--- a/engine/Sandbox.Engine/Systems/Networking/System/NetworkSystem.Handshake.cs
+++ b/engine/Sandbox.Engine/Systems/Networking/System/NetworkSystem.Handshake.cs
@@ -428,6 +428,11 @@ internal partial class NetworkSystem
 		};
 
 		source.SendMessage( output );
+
+		if ( Application.IsEditor )
+		{
+			IToolsDll.Current?.SetPlaying();
+		}
 	}
 
 	Task On_Handshake_ClientReady( ClientReady msg, Connection source, Guid msgId )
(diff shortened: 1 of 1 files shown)


dictateurfou:
that cause some issue with loading scene and snapshot i close the pr i think the original way for make that is bad

PR #11530, opened 2026-07-19
Optimize ComponentList.Get(All)

peter-r-g:
## Summary

This PR optimizes all `ComponentList.Get` and `ComponentList.GetAll` methods.

## Motivation & Context

I first noticed this while looking into performance issues with [Alasi Online](https://sbox.game/duckanddinosaur/alasi-online) and noticed that ~94% of their whole `Scene.Update` was taken by calls to `ComponentList.GetAll`
image

## Implementation Details

This PR does not affect any public API but affects the internal API to `ComponentList` itself. Instead of passing `List<T>`s to the `CollectAll` method it does a lazy enumeration. This means now all `ComponentList.Get` calls will benefit from this optimization since they would fallback to `ComponentList.GetAll` and incur the full iteration cost. `ComponentList.GetAll` would also create a 16 element `List<T>` on every invocation so now getting a single component will not create lots of garbage.

## Checklist

### DIFF
--- a/engine/Sandbox.Engine/Scene/GameObject/ComponentList.cs
+++ b/engine/Sandbox.Engine/Scene/GameObject/ComponentList.cs
@@ -214,15 +214,11 @@ public class ComponentList
 	{
 		if ( go.IsDestroyed ) return Enumerable.Empty<T>();
 
-		var results = new List<T>( 16 );
-
-		CollectAll( results, find );
-
-		return results;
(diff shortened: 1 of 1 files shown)


lolleko:
Do you benchmark this? you are trading list allocations for Enumrable/state machine allocations.

lolleko:
Not merging any optimizations without benchmarks.

PR #11535, opened 2026-07-20
2D sprite sorting: sort layers, sort order, sorting groups, transparency sort axis

sinanata:
## What this adds

Deterministic draw order for 2D sprites, following the model most 2D engines use.

**Sort layers.** Named layers defined per project in Project Settings, under a new Sorting section. Sprites reference a layer rather than a raw depth number, so reordering the project's layer list reorders the scene.

**Sort order.** An integer within a layer, for ordering sprites that share one.

**Sorting groups.** A component that makes a subtree sort as a single unit, so a character assembled from many sprites keeps its internal order while sorting against the rest of the scene as one object.

**Transparency sort axis.** A per camera vector controlling how distance is projected when ordering transparent sprites, which is what makes isometric and 2.5D scenes order correctly.

`SpriteRenderer` gains `IsSorted`, `SortLayer` and `SortOrder`. Sorting is opt in and off by default, so existing scenes render exactly as they do today.

### How it works

Ordering is decided on the GPU. `SpriteDrawPlan` builds the per batch plan on the C# side, `sprite_cs.shader` writes a sort key per sprite, and `sort_cs.shader` sorts them. Sort keys pack layer, order and projected distance so one comparison resolves the whole ordering.

## What this fixes

**Multi edit writing one object's value across a whole selection.** Selecting several objects and editing an embedded resource, sprite or generator field wrote a single target's value to every selected object, losing the others. The cause is in the control widgets, which produced their value from a shared rebuild path rather than per target. `EmbeddedResourceControlWidget`, `EmbeddedSpriteControlWidget` and `ResourceGeneratorControlWidget` now write per target, and rebuild time side effects that could overwrite the selection are suppressed.

This is independent of the sorting work and reproduces without it.

## On the second commit

The second commit also passes buffer capacity into `sprite_cs.shader` and guards the atomic counter trail writes against it. To be clear about what that is: it bounds the sort key buffer that the first commit introduces, so it is internal to this pull request rather than a fix to anything already shipped. Without it a scene with two sorted sprites could lose the second, and three could hang the GPU.

I kept it separate to keep the review diff honest, but I am happy to squash it into the first commit if you would rather review a single clean one.

## Tests

Unit tests covering sort key packing, the draw plan and its indexing:
`SpriteSortKeyTest`, `SpriteDrawPlanTest`, `SpriteDrawIndexingTest`.

Integration tests covering sorting behaviour and the multi edit fix:
`SpriteSortingTests`, `SpriteMultiEditTests`, `EmbeddedSpriteMultiEditTest`, `TextureGeneratorIdentityTest`.

## Notes

There is no proposal issue behind this. Happy to open one and discuss the design first if you would rather do that before spending review time on it.

### DIFF
--- a/engine/Sandbox.Engine/Scene/Components/Camera/CameraComponent.Sorting.cs
+++ b/engine/Sandbox.Engine/Scene/Components/Camera/CameraComponent.Sorting.cs
@@ -1,0 +1,106 @@
+namespace Sandbox;
+
+/// <summary>
+/// How sorted sprites are ordered against each other by a camera, once their sort layer and
+/// order in layer have been compared and come out equal.
+/// </summary>
+[Expose]
+public enum TransparencySortMode
+{
+	/// <summary>
+	/// Follow the camera's projection - distance from the camera when perspective, depth along
+	/// the view direction when orthographic.
+	/// </summary>
+	[Icon( "auto_awesome" )]
+	Default,
+
+	/// <summary>
+	/// Always sort by distance from the camera position, even in an orthographic view.
+	/// </summary>
+	[Icon( "videocam" )]
+	Perspective,
+
+	/// <summary>
+	/// Always sort by depth along the camera's view direction, even in a perspective view.
+	/// </summary>
+	[Icon( "crop_free" )]
+	Orthographic,
+
+	/// <summary>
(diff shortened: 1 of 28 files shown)


antopilo:
I don't think 4k LOC for something that could have been resolved with a simple z-index system is worth trying to fix up. This is doing way too much.

PR #11538, opened 2026-07-20
Add Z Index sorting for sprite

PolEpie:
# Pull Request

---

## Summary

Since i struggled too much with particles sorting order, and i saw a fucking sloppy pr today regarding it, i was like let's try doing a first draft of how i would like it to work

## Motivation & Context

So using particles you sometimes need to layer them on same pos, and want a specific one to show on top of another
So i came up with mutiples idea of doing it, first it was like if 2 particles have the same Z Index, it use distance, but it was a too simple solution with it's own flaws (so if there was 2 particles system with diff pos it would overlap weirdly)

Then i was like i need to take distance in count first and kinda modify it so Z index change the distance, so i made the choice of using a 1 Z Index = 1 inch, and this seem much more cleaner cause it mean moving a particles more in front of camera will show it on top even if it have a lower Z index and that's exactly what we want

Fixes: #10251 

## Implementation Details

So as i said previously we take the distance of sprite as we use to took it before for sorting into sprite_cs, and then we do

> 	// Sort key compared as uint by sort_cs. The Z index biases the sort distance toward the
	// camera, so sprites close to each other layer stably by Z index while clearly separated
	// sprites still sort by real distance. 0xFFFFFFFF is reserved for invalid slots.
	uint CalculateSortKey(float3 worldPosition, int zIndex)
	{
		float biasedDistance = CalculateDistance(worldPosition) - zIndex * SORT_BIAS_PER_Z_INDEX;
		return FloatToSortableUint(biasedDistance);
	}

So that way, both distance and Z index working and a 0 Z index will change nothing and so every particles will still working as expected

Classic Z index priority
https://github.com/user-attachments/assets/d8785ef7-b2ea-4b1b-97ee-8a8bea08187a

Z index along position
https://github.com/user-attachments/assets/6b1a4405-5c2f-4197-b44c-697765e12d2b

## Checklist

### DIFF
--- a/engine/Sandbox.Engine/Scene/Components/Particles/Renderers/IBatchedParticleSpriteRenderer.cs
+++ b/engine/Sandbox.Engine/Scene/Components/Particles/Renderers/IBatchedParticleSpriteRenderer.cs
@@ -35,6 +35,7 @@ internal interface IBatchedParticleSpriteRenderer : ISpriteRenderGroup
 	float CameraFadeFar { get; }
 	float FogStrength { get; }
 	FilterMode TextureFilter { get; }
+	int ZIndex => 0;
 
 	// Implemented by derived classes
 	Texture RenderTexture { get; }
@@ -88,6 +89,7 @@ internal interface IBatchedParticleSpriteRenderer : ISpriteRenderGroup
 		var packedFogAndAlpha = SpriteData.PackFogAndAlphaCutout( this.FogStrength, 0.001f );
 		var depthFeather = DepthFeather;
 		var packedCameraFade = SpriteData.PackCameraFade( CameraFadeNear, CameraFadeFar );
+		var zIndex = ZIndex;
 		var blurOpacity = BlurOpacity;
 		var origin = Pivot;
 		var renderFlags = SpriteFlags.None;
@@ -193,6 +195,7 @@ internal interface IBatchedParticleSpriteRenderer : ISpriteRenderGroup
 				spritePtr->SequenceTime = sequenceTime;
 				spritePtr->BlendSheetUV = sequenceData;
 				spritePtr->Offset = origin;
+				spritePtr->ZIndex = zIndex;
 
 				validCount++;
 			}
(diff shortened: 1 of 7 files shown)


CarsonKompon:
This looks nice, clean, and has a relatively small footprint.
Would love a second opinion from @antopilo before triaging

antopilo:
@CarsonKompon I don't like that z-index is additive to distance from camera. What if you have a sprite that is further but want to display above everything else? you'd need a huge z-index to compensate for the camera distance?

If sprite A has z-index 1 and other sprites have z-index 0. sprite A should be above everything disregarding depth? that's my intuition 🤔 Maybe im wrong? @ryleigh what do you think?

PolEpie:
Okay I get you point but that’s not really for our usecase, mine is more for like localized Z index so 

if I have my GO with 2 / 3 layers, they are all sprite renderer with the same pos
I wanna be able to decide the order of each ones to layer them,
But I don’t wanna specifically say let’s render this red effect on top of every sprite

for the mode you said, it would be interesting to add the Z index to not sorted mode 
So only Z index is working here, no distance calculation but I don’t see a use case for that mode tbh

PolEpie:
Tbh i starting to get your point @antopilo but the thing is for effects, i'm scared to have a weird layering especially with all transparent layers

Here is a exemple before the changes your talked about
(changes in sprite_cs)

```
// Sort key compared as uint by sort_cs. Upper 8 bits hold the Z index layer, lower 24 bits
// hold the camera distance: Z index dominates, so a higher Z index always draws on top
// regardless of distance, and distance sorts back-to-front within the same layer.
// 0xFFFFFFFF is reserved for invalid slots (sorts past any valid key).
uint CalculateSortKey(float3 worldPosition, int zIndex)
{
  // Inverted so a higher Z index produces a smaller key, which draws later (on top)
  uint zPart = (uint)(127 - clamp(zIndex, -128, 127));
  uint distPart = FloatToSortableUint(CalculateDistance(worldPosition)) >> 8;
  return (zPart << 24) | distPart;
}
```

https://github.com/user-attachments/assets/3a63b991-c8ac-4837-805f-e1a0c30885d0

Here is now

https://github.com/user-attachments/assets/99c996c6-90df-46e5-aebe-861da968c6ca

The result is more whitish cause white have a 1 z index
But maybe it's the best way of tackling this issue

PolEpie:
Better exemple

With the change of Z index layer anto asked
image

Without (like the pr)
image


it's super visible on that part, it's a detail but just wanna be sure it's taken in consideration
image

PolEpie:
Btw all the changes needed has been done on my side, it’s now based on the z index and use dist if 2 sprites have the same index :)

PR #11539, opened 2026-07-20
Add localization support for `menu` addon

zxcPandora:
## Summary
This PR introduces comprehensive localization support to the Menu UI. Hardcoded strings across various menu components  in the `menu` addon have been extracted and replaced with standardized localization keys.

## Motivation & Context
Previously, text within the Menu UI in the `menu` addon was hardcoded in English. 
By extracting these strings and mapping them to localization keys, it allows the community to easily contribute translations for their respective languages without needing to touch the UI source code.

## Implementation Details
- Extracted hardcoded UI text in the `menu` addon and replaced them with localization keys.
~~- **Primary Method (`MenuHelpers.cs`)**: Added new `Localize` function in `MenuHelpers.cs` to handle the majority of string replacements. Proper `<summary>` tags have been included to document its usage for future contributors.~~
- **Markup / Prefix Method**: Utilized the `#` prefix approach for localization directly within certain UI components.
- **Special Cases**: A very small number of edge cases required special localization handling, including:
  - `game\addons\menu\Code\DevUI\DevMode\DeveloperMode.razor`
  - `game\addons\menu\Code\MenuUI\Components\Multiplayer\GameGroupRow.razor`
  - `game\addons\menu\Code\MenuUI\Components\PackageFilters\PackageCategoryFilter.razor*`
- Ensured UI placeholder orders and string interpolations (e.g., `{count}`, `{name}`) are properly formatted to support different grammatical structures in various languages.

## Checklist


zxcPandora:
### Quick Q&A / Notes
To keep the main PR description clean, I’ve added a quick Q&A here to address some potential questions regarding this submission:

**Q: Why does this PR only include English and Simplified Chinese localization? Why not use AI to generate translations for other languages?**
**A:** 
1. Due to the game's existing code design, any missing localized strings will automatically fall back to English anyway.
2. My native language is Chinese, and my second language is English. I haven't studied other languages. 
3. Because I lack the cultural context and grammatical knowledge of other regions, I cannot accurately evaluate or guarantee the quality of AI-generated translations. I prefer to leave this to native speakers in the community.

~~**Q: The project already has the built-in `Game.Language.GetPhrase` method. Why did you need to add a new custom function?**~~
~~1. When the built-in method encounters an error or a missing translation, it simply returns the format string. The custom function introduces parameters to give us better control over whether to return the raw `key` as a fallback.
2. The custom function centralizes string formatting. This keeps the localization implementation concise and highly readable, preventing UI files from being cluttered with excessively long function calls and repetitive code.~~

**Q: How can I test this PR on my end?**
**A:** 
Clone the project to your local machine, run `Bootstrap.bat`, and wait for it to complete. Make sure Steam is running in the background, and then launch the s&box main executable located in the `game` directory.
*(Note: If, like me, you are compiling from this public repository without owning s&box on Steam yet, please modify the `AppId` variable in `engine\Sandbox.Engine\Application.cs` to `480` and save it **before** running `Bootstrap.bat`. This is the free AppID for Spacewar, which allows you to test Steamworks SDK integration without an actual AppID.)*

**Q: I compiled and tested your PR, but I still see some unlocalized text. Why is that? (And how does this relate to the formatting?)**
**A:** As mentioned in the PR title, this localization work is strictly scoped to the `menu` addon. The remaining unlocalized strings you might see generally come from:
1. Hardcoded strings deep within the engine's core code.
2. Strings originating from the `base` addon.
3. A very small portion of text fetched directly from the s&box backend/website. 
*(Note: Dealing with this mixed environment of localized and unlocalized strings from different sources is actually another reason why the new custom function was necessary to handle formatting safely and gracefully.)*

**Q: What naming convention did you use for the new localization keys?**
**A:** The keys follow a hierarchical naming structure to keep things organized and traceable:
- **Standard (Majority):** `FileName.PrimaryKey`
- **Nested:** `FileName.ParentObject.PrimaryKey`
- **Deeply Nested:** `FileName.ParentObject.ChildObject.PrimaryKey`
- **Tooltips:** For any tooltip text, I simply prepended `tooltip.` to the structures mentioned above (e.g., `tooltip.FileName.PrimaryKey`).

**Q: Why are relative timestamps (like "1 min ago" or "3 days ago") still showing up in English?**
**A:** Those relative time strings are automatically generated by a third-party .NET library used by the engine called **Humanizer** (`Humanizer.dll`). While the library itself supports multiple languages, it requires its specific localized resource assemblies (e.g., satellite DLLs for `zh-CN` or other locales) to be shipped alongside the engine. 
Since this PR strictly focuses on the UI layer, I cannot inject engine-level dependencies. The localization of these specific time strings will naturally be resolved once Facepunch decides to include the localized resource files for `Humanizer` in a future engine update.


Any feedback, suggestions, or guidance is highly welcome!

game/addons/menu/Code/MenuUI/Components/GameStarting.razor
```diff
@@ -14,7 +14,7 @@
 
         <div class="game-title">@GameTitle</div>
         <div class="game-author">@GameAuthor</div>
-        <div class="more-info"><span class="button">F1</span> for info, hold <span class="button">ESC</span> to exit</div>
+        <div class="more-info"><span class="button">F1</span>@MenuHelpers.Localize("gamestarting", "prefix")<span class="button">ESC</span>@MenuHelpers.Localize("gamestarting", "suffix")</div>
```
CarsonKompon: We should never be using prefix/suffix strings, as all languages work differently.
Instead, you should look at the [Advanced Tokens](https://sbox.game/dev/doc/ui/localization) example in the documentation, and do something like this:
##### en/menu.json:
```json
"gamestarting.phrase": "{Button1} for info, hold {Button2} to exit.",
```
##### GameStarting.razor:
```cs
var data = new Dictionary<string, object>();
data["Button1"] = "F1";
data["Button2"] = "ESC";
var message = Language.GetPhrase( "gamestarting.phrase", data );
```
OR
```cs
Language.GetPhrase( "gamestarting.phrase", new() { ["Button1"] = "F1", ["Button2"] = "ESC" } );
```

game/addons/menu/Code/MenuHelpers.cs
```diff
@@ -271,4 +271,43 @@ public static void LaunchGame( string gameIdent, bool allowLaunchOverride = true
 
 		Game.Overlay.ShowGameModal( gameIdent );
 	}
+
+	/// <summary>
+	/// Look up a localized phrase by fullKey.
+	/// When <paramref name="returnOriginKey"/> is true, returns <paramref name="fullKey"/>
+	/// untouched instead of the full key on lookup failure.
+	/// </summary>
+	public static string Localize( string fullKey, bool returnOriginKey = false )
```
CarsonKompon: We definitely don't need `MenuHelpers.Localize` functions.

Lets just modify `Language.GetPhrase()` so that they return the english text instead of the token if the token isn't present in the current language... Since `Language.GetPhrase` can already do everything else you need.

game/addons/menu/Code/MenuHelpers.cs
```diff
@@ -271,4 +271,43 @@ public static void LaunchGame( string gameIdent, bool allowLaunchOverride = true
...
+		var result = Game.Language.GetPhrase( normalized, data );
+		return returnOriginKey && result == normalized ? fullKey : result;
+	}
+
+	/// <summary>
+	/// Look up a localized phrase with a prefix appended to the key, plus optional template data.
+	/// When <paramref name="returnOriginKey"/> is true, returns <paramref name="key"/>
+	/// untouched instead of the full prefixed key on lookup failure.
+	/// </summary>
+	public static string Localize( string prefix, string key, Dictionary<string, object> data = null, bool returnOriginKey = false )
```
CarsonKompon: Why does this one even exist? Seems really silly when you just need to concat strings to do the same thing.

As mentioned before, remove all the `MenuHelpers.Localize()` functions and make it so you use `Language.GetPhrase` everywhere.

zxcPandora: right,let me have a check in the engine and then change all.thanks for the advice

zxcPandora: > Why does this one even exist? Seems really silly when you just need to concat strings to do the same thing.
> 
> As mentioned before, remove all the `MenuHelpers.Localize()` functions and make it so you use `Language.GetPhrase` everywhere.

I am a bit undecided right now, I came up with two methods (for handling those passed-in values that might contain special symbols, such as `" "`,`"#"`, `"-"`)

 - Method 1 (only handle those without special strings in `engine`, and handle them by themselves in the files containing special strings):
```csharp
public string GetPhrase( string phrase, Dictionary<string, object> data = null, bool returnSegment = false )
{
	if ( !Phrases.TryGetValue( phrase, out var result ) )
	{
		if ( returnSegment )
		{
			var lastDotIndex = phrase.LastIndexOf( '.' );
			return lastDotIndex > 0 ? phrase[(lastDotIndex + 1)..] : phrase;
		}
		return phrase;
	}

	return result.Render( data );
}
```

For those requiring special character handling, add a custom function in the passed-in file:
```csharp
string Localize( string originalText )
{
	// Directly replace special strings with empty or replace with underscores
	string localizedText = Language.GetPhrase( "example." + originalText.Replace( " ", "_" ).Replace( "#", "_" ).Replace( "-", "_" ) );
	if ( string.IsNullOrEmpty( localizedText ) || localizedText.StartsWith( "example." ) )
		return originalText;

	return localizedText;
}
```

 - Method 2 (do not handle in the files containing special strings, uniformly hand over to `engine` for handling):
```csharp
public string GetPhrase( string phrase, Dictionary<string, object> data = null, bool returnSegment = false )
{
	// Exact match first
	if ( Phrases.TryGetValue( phrase, out var result ) )
		return result.Render( data );

	// Search again after formatting (replace special strings with empty or replace with underscores)
	var normalized = phrase.Replace( " ", "" ).Replace( "#", "" ).Replace( "-", "" );
	if ( normalized != phrase && Phrases.TryGetValue( normalized, out result ) )
		return result.Render( data );

	// Cannot find either: use the original key to get the last segment
	if ( returnSegment )
	{
		var lastDot = phrase.LastIndexOf( '.' );
		return lastDot > 0 ? phrase[(lastDot + 1)..] : phrase;
	}
	return phrase;
}
```

game/addons/menu/Code/MenuUI/Components/GameStarting.razor
zxcPandora: Hi! I tried the Advanced Tokens approach for gamestarting.phrase, but the F1/ESC keycaps lose their styling. They're currently <span class="button"> elements (white bg, rounded, padded, shadow via SCSS), and rich-text spans only support text-level styles (font/color/background) —padding, border-radius and box-shadow don't render. Any recommended way to keep styled keycaps inside a localized phrase, or should I just use plain text here?

zxcPandora:
**Quick question about path casing.** I noticed `game/addons/menu/code/` uses lowercase in some files (e.g. `code/Modals/TagSelector.razor`) while most of the repo uses `Code/` (capital). I've been fixing these lowercase paths to `Code/` in my branch, but it shows up as a bunch of "renamed without changes" entries in the PR.

Would you prefer I keep these casing fixes, or should I revert them to keep the diff clean?

PR #11549, opened 2026-07-22
Editor play as client fix

dictateurfou:
## Summary

Fixes editor play mode when connecting to a dedicated server.

The editor previously displayed the correct base scene, but it entered play mode through the normal local `EditorScene.Play( true )` path. This starts the game again instead of presenting the scene that was reconstructed from the server snapshot. Consequently, objects created at runtime and included in the snapshot were not reliably visible in the editor viewport.

The change adds a dedicated client activation path which presents the fully loaded network scene without restarting the game.

## Motivation & Context

The editor must enter play mode only once the client has completed the normal connection and snapshot-loading flow.

Previously, editor play activation was triggered from handshake and scene-load message handlers. This could happen while the snapshot was still being applied, allowing play-mode side effects and RPCs to run before the player had finished loading. It also reused the regular local-play flow, which is not appropriate for an already connected client receiving an authoritative scene snapshot.

Fixes: N/A

## Implementation Details

- Replaces the ambiguous `SetPlaying` tools API with `PlayAsClient`, making this editor-client flow explicit.
- Moves the editor activation to the end of `SceneNetworkSystem.SetSnapshotAsync`, after:
  - the scene has finished loading;
  - snapshot data, including runtime-created network objects, has been applied;
  - `SceneLoaded` has been signalled;
  - client initialization has completed.
- Restricts the path to editor clients with `Application.IsEditor && Networking.IsClient`.
- Adds `EditorScene.PlayAsClient()`, which attaches the loaded `Game.ActiveScene` to the editor session without calling `Play( true )` or `EditorPlay()` again.
- Stops any existing playing editor session before attaching the loaded scene, avoiding stale or duplicate sessions.
- Removes editor activation from handshake and scene-load message handlers. The network layer now remains focused on handshake and snapshot processing, while the editor transition is centralized at the existing snapshot-completion point.

### Before

> The correct base scene is displayed, but runtime objects received through the server snapshot are missing.

whithout_fix

### After

> The editor displays the fully reconstructed snapshot, including runtime-created objects.

withfix

## Checklist

- [x] Public APIs are documented


engine/Sandbox.Tools/Scene/EditorScene.cs
```diff
@@ -298,6 +298,29 @@ public static void Play( bool playMode, SceneEditorSession playableSession = nul
 		EditorEvent.Run( "scene.play" );
 	}
 
+	/// <summary>
+	/// Starts the scene once loading is complete when using Connect command in the editor.
+	/// </summary>
+	public static void PlayAsClient()
+	{
+		var sceneEditorSession = SceneEditorSession.All.FirstOrDefault( x => x.IsPlaying );
+		sceneEditorSession?.StopPlaying();
```
deadmonstor:
```suggestion
		SceneEditorSession.Playing?.StopPlaying();
```

We can do this now that we are inside of EditorScene, probably more stable for the future

engine/Sandbox.Engine/Scene/Networking/SceneNetworkSystem.cs
```diff
@@ -658,6 +653,12 @@ public override async Task SetSnapshotAsync( SnapshotMsg msg )
 		}
 
 		Game.IsPlaying = true;
+
+		// used when connecting from the editor to display the game.
+		if ( Application.IsEditor && Networking.IsClient )
+		{
+			IToolsDll.Current.PlayAsClient();
```
deadmonstor:
```suggestion
			IToolsDll.Current?.PlayAsClient();
```

Just incase, everywhere else does it 😆

dictateurfou: It’s because I check Application.IsEditor, so IToolsDll is guaranteed to exist. The places using IToolsDll.Current? do not check whether they are running in the editor, so the null-conditional access partly serves as that guard. I could probably replace the whole block with just that.

PR #11552, opened 2026-07-23
Add a test which makes sure the results from Game.Random are stable forever

nixxquality:
## Summary

This PR just adds a simple unit test as a tripwire for the randomization API.

## Motivation & Context

Randomization is used for lots of reasons in games. One case in my mind here is random level generation. You want an algorithm which does the same thing forever.
This unit test makes sure we'll find out if something is going to break the randomization algorithm developers might be relying on.

## Checklist


lolleko:
why would something break the randomization?

nixxquality:
I don't know. There are plenty of things that break for no reason.

In my eyes, if there's no test which guarantees a behaviour, it's not something you can rely on.

PR #11590, opened 2026-07-31
Faster Transform.Lerp Transform.ToLocal and constructor

boxrocket6803:
## Summary
`Transform.Lerp` and `Transform.ToLocal` as well as the constructor were all a bit more expensive than they usually need to be.

## Motivation & Context
Easier than optimizing my animation system lol

## Implementation Details
The Transform constructor previously used the shorthand `Scale = 1.0f`. Using `Scale = Vector3.One` instead saves having to go through a stack of `Vector3` constructors every time a transform is created

`Transform.ToLocal` does a lot of fairly complex stuff to handle scale, but I find that the vast majority of time I use it the scale on both transforms is set to 1. Under that assumption it now skips the scale calculation if both scales equal 1, making it roughly equal to `ToWorld` in terms of performance.

I tend to leave `Transform.Lerp` sitting at 0 or 1, adding specific checks for that saves a lot of time in my code, perhaps it will for others as well.

## Screenshots 
image
image

## Checklist

### DIFF
--- a/engine/Sandbox.System/Math/Transform.cs
+++ b/engine/Sandbox.System/Math/Transform.cs
@@ -61,7 +61,7 @@ public struct Transform : System.IEquatable<Transform>, IInterpolator<Transform>
 	{
 		Position = pos;
 		Rotation = Rotation.Identity;
-		Scale = 1.0f;
+		Scale = Vector3.One;
 	}
 
 	public Transform() : this( default )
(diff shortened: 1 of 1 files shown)


lolleko:
Would prefer proper benchmarks, rather than a profiler screenshot.
We already have a micro benchmark harness, you can modify to run your test.

Do not only measure the good cases (stuff left at 0 or 1) also measure with other numbers, ideally just throw random numbers in there.

boxrocket6803:
benchmark revealed my edits to `ToLocal` didn't actually make it any faster (or more specifically it's faster in debug but not on release), but I discovered that using `SafeScale` for both position and scale as opposed to the `localSafeScale` contraption that previously existed makes a decent improvement in both modes.

`LerpTo` worked exactly as expected. Using `Vector3.One` in the constructor makes no difference at all on release, but since it's a bit faster in debug it might as well stay I feel.

The following are run with some measures in place to stop the compiler from optimizing everything out for the `Lerp` methods (which accounts for a bit of extra time on both). The `Lerp` method also includes 2 calls at different values
```
existing sbox code
| Method         | Mean      | Error     | StdDev    | Median    |
|--------------- |----------:|----------:|----------:|----------:|
| LerpPinned     | 85.587 ns | 0.2950 ns | 0.4231 ns | 85.527 ns |
| LerpRandom     | 80.410 ns | 0.2324 ns | 0.3479 ns | 80.484 ns |
| ToLocalNoScale | 13.314 ns | 0.1034 ns | 0.1483 ns | 13.278 ns |
| ToLocalScaled  | 13.325 ns | 0.0683 ns | 0.1001 ns | 13.305 ns |
```
```
updated code
| Method         | Mean      | Error     | StdDev    | Median    |
|--------------- |----------:|----------:|----------:|----------:|
| LerpPinned     |  3.173 ns | 0.0663 ns | 0.0951 ns |  3.158 ns |
| LerpRandom     | 78.128 ns | 0.6279 ns | 0.9399 ns | 77.804 ns |
| ToLocalNoScale | 11.577 ns | 0.0916 ns | 0.1314 ns | 11.610 ns |
| ToLocalScaled  | 11.608 ns | 0.0507 ns | 0.0727 ns | 11.608 ns |
```

LerpPinned evaluates `LerpTo` with 0 and 1. With existing code it's fairly significantly more expensive than random values, I'm assuming this is due to clamping the value per individual component instead of just once. It's extremely cheap with the new code.
LerpRandom evaluates `LerpTo` with the arbitrary numbers 0.3 and 0.7. Roughly the same performance.

ToLocalScaled evaluates `ToLocal` with two random (but not randomized) transforms.
ToLocalNoScale evaluates `ToLocal` using the same transforms with their scale is set to 1.
All the transforms are created outside the benchmark method to avoid changing the results. Only needing to create 1 new Vector3 instead of 2 (as well as only one set of conditional expressions) seems to make a pretty decent difference.

All adds up to a fairly significant improvement in performance in my custom animation system, which relies heavily on both `Lerp` and `ToLocal`

Issue #11630, opened 2026-08-07
Modifying a big prefab into a scene is unusable


PolEpie:
### Version / Commit

26-08-05

### Describe the bug

So i have a big prefab of a part of my map, and i put it on my main scene to start working on it, but modifying it directly from the scene it's put on it's not possible, a simple action that have a GameObjectSnapshot with it for undo gonna take 2/3s, it's unusable

### To Reproduce

Take a big prefab full of Scene mapping

### Expected behavior

Just act like hammer does with prefab, so it's flawlessly

### Media/Files

<img width="2891" height="1595" alt="Image" src="https://github.com/user-attachments/assets/de53ebeb-b201-4ba2-b9a1-4c666b699114" />

https://github.com/user-attachments/assets/545f8994-fae1-4c14-8d9d-e8a8f4db0542

PR #11631, opened 2026-08-07
Avoid rebuild old FindTrackedObjectsInJson into json CalculateDiff

PolEpie:
# Pull Request

---

## Summary / Motivation & Context

This is part of a big finding on my #11630 issue
So i saw that FindTrackedObjectsInJson was called twice and did it with old / new, and i was like hmm, couldnt that being cached ?
image

## Implementation Details

So my pr is simple, into PrefabCacheScene.Load, we calculate the Json.BuildTrackedObjects directly, and use it on FindTrackedObjectsInJson instead of recalculate it.
Anyway once a prefab is saved it's gonna recall load and update it

## Checklist


lolleko:
Good idea, but i don't like the fact though that we are making diffing internals public/internal.
Will think about it.

PR #11632, opened 2026-08-08
Optimisation of Scene undo when working with prefab

PolEpie:
# Pull Request

---

## Summary - Motivation & Context

When doing any action with the scene mapping that involve a undo action, it was taking 2-3s, and i profile and discover that it would serialize twice for doing 1 action

(there is only 1 action done, you see the 2 part with serialization)
image

## Motivation & Context
 
I use only scene mapping, and one of the biggest complain of my mappers when using it with prefab was #11630 
So i dig a bit using SuperLuminal and my friend JeanClaude and i discover that it would Serialize the prefab on ctor and dispose, so 2 time serialization + i think we could do it better for sure

This pr is made with help of ai but i did reread it and verify it myself to not doing shit, i dont have the holytruth but i tried it as i can and seem to work, might be good tackle some of these issues :)

## Implementation Details

Basicly i tried to break it down into multiples commit and limit the footprint:

- First i discover that there was a issue where it tried to access a `go.Parent` and `go.IsPrefabInstance` before the `!go.IsValid() || go.IsDestroyed`

- Then there was a duplicate of RefreshPatch when snapshotting prefab instance root, cause `GameObjectSnapshot` called `RefreshPatch()` explicitly right before `go.Serialize()`,
which already calls it internally via `SerializePrefabInstance`. Each call serializes and
diffs the whole instance, so every snapshot of an instance root paid that cost twice.

- Then it was the bigger and most important part, that was why it snapshotted the entire instance twice when undo due to `_initalCapturedGameObjects[OutermostPrefabInstanceRoot] = All`, now :

  - Edited objects/components are captured directly with their declared flags; the
  instance root's id is collected instead (by id — undo/redo can recreate objects).
  - `RefreshPrefabPatches` recomputes the cached patch of collected roots once at scope
  close. This keeps the contract the promotion satisfied implicitly: unpatched instance
  changes are discarded when the prefab refreshes, and the override indicators read it.
  - The undo/redo actions refresh the same roots after restoring, after the callback batch
  has flushed — deferred `PostDeserialize` rebuilds blob-backed component state (meshes)
  that patch serialization reads (hence the redo batch becoming block-scoped).
  - Roots still captured in patch form (flags include `Children`, e.g. deletions) already
  refresh during serialization and are skipped.

- Finally i made so if i simply move something, it will forces the standard serialization path via
`SerializePrefabForDiff`, storing just the root's own properties instead of the patch
form and its full-instance `RefreshPatch`.

Before:
https://github.com/user-attachments/assets/ee104553-dc9c-460a-940d-b5a1c97ee3f9

After:
https://github.com/user-attachments/assets/673c72f8-f8b1-433f-b18a-b17af45ce5db

With that commit, now i can move easily the props i have, but it will freeze on dispose (around 1s time to serialize everything + my pr #11631 can help reduice the time), anyway my map is big so i'd except a 1s freeze sometimes, but i prefer it after release mouse / doing action, rather than on ctor when pressing mouse

Before (1 translation):
image

After (2 translation):
image

## Checklist

### DIFF
--- a/engine/Sandbox.Tools/Scene/Session/SceneEditorSession.Undo.cs
+++ b/engine/Sandbox.Tools/Scene/Session/SceneEditorSession.Undo.cs
@@ -290,7 +290,6 @@ internal sealed class SceneUndoSnapshot : IDisposable
 				}
 				var serializeOptions = new GameObject.SerializeOptions { IgnoreChildren = !flags.HasFlag( GameObjectUndoFlags.Children ), IgnoreComponents = !flags.HasFlag( GameObjectUndoFlags.Components ) };
 				GameObjectRefs.Add( GameObjectReference.FromInstance( go ) );
-				if ( go.IsOutermostPrefabInstanceRoot ) go.PrefabInstance.RefreshPatch();
 
 				using var blobs = BlobDataSerializer.Capture();
 
--- a/engine/Sandbox.Tools/Scene/Session/SceneEditorSession.Undo.cs
+++ b/engine/Sandbox.Tools/Scene/Session/SceneEditorSession.Undo.cs
@@ -1,4 +1,4 @@
-﻿using Sandbox.Helpers;
+using Sandbox.Helpers;
 using System;
 using System.Text.Json.Nodes;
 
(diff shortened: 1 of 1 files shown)


lolleko (commented):
Scary. Will have a closer look might end up doing this ourselves I think this could be less hacky,

Really good analysis, though.

PolEpie:
@lolleko if you want access to my project with the huge prefab lmk :)
Might be better for you to have it to debug

And if you have any question shoot me a dm on discord (I'm the pol without hardhat on my name)

PR #11638, opened 2026-08-10
Add open_scene MCP tool to switch/open scene tabs

### LINKED ISSUE #11637: Proposal: MCP tool to switch/open scene tabs (open_scene)
The `scene` toolset can read any open scene, but nothing can change which one is active, or open a `.scene` that isn't open at all. The `scene` parameter is read-only scoping: `set_component`, `set_game_object`, `save_scene` and `undo` all implicitly hit the active session.

That combination silently loses edits. `FindByGuid` walks every open session and returns the first match, so `set_component` aimed at a component in a background tab does mutate that object - but the dirty flag lands on the active tab, `save_scene` saves the active scene, and the background scene reloads from disk on focus. With duplicated scenes (same component guids in two files) the agent can't even tell which one it hit. Today the only fix is asking a human to click the right tab, which defeats the point of driving the editor over MCP.

Both behaviours already exist as public API:

```csharp
/// <summary>
/// Make a scene the active editor tab. Opens it from its asset path if it isn't open.
/// Scene edits always target the active scene, so switch before editing a background scene.
/// </summary>
/// <param name="scene">Scene name or resource path from list_scenes, or a .scene/.prefab path from asset_search.</param>
[McpTool( "open_scene" )]
public static OpenScene OpenScene( string scene )
```

backed by `SceneEditorSession.MakeActive()` (the same call `SceneTabWidget` makes on click) and `SceneEditorSession.CreateFromPath( path )`, which already returns the existing session if the scene is open. It's a `ResolveSession` helper mirroring `ResolveScene` (Scene.cs ~1002), a short tool method and a small result DTO.

Notes:

- Refuse `GameEditorSession` - it has no tab, so `MakeActive` half-switches the editor invisibly. Error points at `play_stop`, guarded like the `play` toolset guards.
- No undo step: tab focus is editor UI state, not scene state. Worth saying out loud since the server instructions promise undo on scene edits, and switching changes what a later `undo` targets.
- Plain `[McpTool]`, not `[McpTool.ReadOnly]`.
- Not-found error points at `list_scenes` / `asset_search`, matching the existing idiom.
- Name is bikesheddable - `open_scene` covers both open-and-activate, but tool names are permanent public API, so I'll take whatever you prefer.

Happy to PR it if the shape looks right - it'd be about the size of #11557.

kitsupanic:
## Summary

Adds `open_scene` to the scene MCP toolset — makes a scene the active editor tab, opening it from its asset path first if it isn't open. Fills the gap where MCP clients could read background scenes but never change which scene is active, and mutating tools (set_component, save_scene, undo) always target the active session.

## Motivation & Context

Fixes: #11637

MCP clients could read and list background scenes, but edits aimed at them silently lose data. FindByGuid takes the first match across all open sessions, so `set_component` aimed at a background tab's component does mutate that object — but the dirty flag lands on the active tab, `save_scene` saves the active scene, and the background scene reloads from disk on focus, discarding the change with no error anywhere. The only workaround was asking a human to click the tab. This tool closes that gap by letting MCP clients make the right scene active before editing it.

## Implementation Details

- Built on existing public API: `SceneEditorSession.MakeActive()` (the same call SceneTabWidget makes on tab click) and `SceneEditorSession.CreateFromPath()`, which already returns the existing session when the scene is open.
- Method is named `OpenSceneTab` because SceneTools already nests the `OpenScene` DTO; a second public method would trigger CS0102. The tool name in the MCP schema is still `open_scene`.
- No undo step, deliberately: tab focus is editor UI state, not scene state. The schema description warns that scene edits target the active scene, so callers know the contract.
- Refuses during play with a message pointing at `play_stop`. A second guard refuses `GameEditorSession` specifically — it has no tab, so `MakeActive` would half-switch the editor invisibly. In practice the play guard fires first, making this defense-in-depth for the transient state where a game session exists while `IsPlaying` is false.
- Two small shared helpers extracted: the existing `ResolveScene` now delegates to a new `ResolveSession` (its error message unchanged), and `list_scenes` row construction moved to a `SceneRow` helper both tools now use.
- Verified in a source-built editor over MCP: registration and schema, open-from-disk, tab switch by name and by path (viewport follows), no-op on already-active, both error paths returning proper `isError` results with no console exceptions, play-mode guard, and regressions on `list_scenes` shape and `scene_tree`'s explicit-scene argument including its exact not-found message. Manual Ctrl+Z after eight `open_scene` calls confirmed no undo steps pushed.

This is a headless-visible change — tool results and tab focus in the editor. No screenshots apply.

## Checklist

  - The MCP tool layer has no unit test harness; verification was manual over MCP as described above.

### DIFF
--- a/game/addons/tools/Code/Mcp/Scene.cs
+++ b/game/addons/tools/Code/Mcp/Scene.cs
@@ -18,16 +18,32 @@ public static partial class SceneTools
 	{
 		return new OpenSceneList
 		{
-			Scenes = SceneEditorSession.All.Select( session => new OpenScene
-			{
-				Name = session.Scene?.Name,
-				ResourcePath = session.Scene?.Source?.ResourcePath,
-				Type = session is GameEditorSession ? "Game" : session.Scene is PrefabScene ? "Prefab" : "Scene",
-				IsActive = session == SceneEditorSession.Active,
-				HasUnsavedChanges = session.HasUnsavedChanges,
-				RootObjectCount = session.Scene?.Children.Count ?? 0
-			} ).ToArray()
+			Scenes = SceneEditorSession.All.Select( SceneRow ).ToArray()
 		};
+	}
+
+	/// <summary>
+	/// Make a scene the active editor tab. Opens it from its asset path if it isn't open.
+	/// Scene edits always target the active scene, so switch before editing a background scene.
+	/// </summary>
+	/// <param name="scene">Scene name or resource path from list_scenes, or a .scene/.prefab path from asset_search.</param>
+	[McpTool( "open_scene" )]
+	public static OpenScene OpenSceneTab( string scene )
+	{
+		if ( Game.IsPlaying )
+			throw new Exception( "Can't switch scene tabs while playing - play_stop first" );
+
+		var session = ResolveSession( scene );
+
+		if ( session is GameEditorSession )
+			throw new Exception( "That's the running game session, which has no tab to switch to - play_stop first, then open the scene you want to edit" );
+
+		session ??= SceneEditorSession.CreateFromPath( scene )
+			?? throw new Exception( $"Nothing to open for '{scene}' - list_scenes shows what's already open, asset_search type:scene finds scene assets on disk" );
+
+		session.MakeActive();
+
+		return SceneRow( session );
 	}
 
 	/// <summary>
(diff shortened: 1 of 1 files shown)


kitsupanic:
Follow-up validation, since the refactor touches shared helpers: ran the full set of mutating tools against a source build with this PR applied — create_game_object, add_component (with properties), set_component, set_game_object, spawn_model, delete_game_object, and undo/redo (undo restored a deleted object with both components intact). All work, clean console throughout.

Also specifically exercised the two read paths that actually go through the refactored code: camera_screenshot with no camera argument and scene_trace both resolve the scene via the ResolveScene("") empty-string branch — both fine. The mutating tools resolve through ActiveSession(), which this diff doesn't touch.

PR #11640, opened 2026-08-10
Fix editor_status reporting the game scene instead of the active editor tab

### LINKED ISSUE #11639: MCP editor_status reports wrong ActiveScene (reads Game.ActiveScene, not SceneEditorSession.Active)
## Version / Commit

26.08.05 (also reproduces on a source build at master)

## Describe the bug

`editor_status` reports the wrong active scene. `TopLevelTools.GetEditorStatus()` reads `Game.ActiveScene` for `ActiveScene`/`ActiveScenePath`:

```
engine/Sandbox.Tools/Mcp/TopLevelTools.cs:146   var scene = Game.ActiveScene;
engine/Sandbox.Tools/Mcp/TopLevelTools.cs:155   ActiveScene = scene?.Name,
engine/Sandbox.Tools/Mcp/TopLevelTools.cs:156   ActiveScenePath = scene?.Source?.ResourcePath,
engine/Sandbox.Tools/Mcp/TopLevelTools.cs:157   SceneHasUnsavedChanges = SceneEditorSession.Active?.HasUnsavedChanges ?? false,
```

The editor's actual active tab is `SceneEditorSession.Active` — which line 157 already uses, and which every scene tool resolves through. `Game.ActiveScene` is the game scene handle, so these two fields report the play-mode scene at best and a stub at worst. One struct, two sources of truth.

## To Reproduce

1. Open two scenes in the editor so both are tabs, e.g. `menu` and `aquarium`. Click the `menu` tab.
2. Call `editor_status` over MCP → `ActiveScene: "Scene"`, `ActiveScenePath: null`.
3. Call `list_scenes` → correctly reports `menu` with `IsActive: true`.
4. Click the `aquarium` tab (or `open_scene` it) and call `editor_status` again → unchanged.
5. Enter and leave play mode → `ActiveScene: null`.

Throughout, `SceneHasUnsavedChanges` in the same response does follow the active tab.

## Expected behavior

`ActiveScene`/`ActiveScenePath` name the scene tab the editor is on — the one `list_scenes` flags `IsActive`.

One-line fix, the same idiom `SceneTools.ResolveScene` uses for its empty-string case (`game/addons/tools/Code/Mcp/Scene.cs:1020-1023`):

```cs
var scene = SceneEditorSession.Active?.Scene ?? Game.ActiveScene;
```

Happy to PR it.

## Additional context

`editor_status`'s own description points clients at it for "the active scene and whether it has unsaved changes", so it's the documented first call for any MCP client — and it's wrong about half of what it advertises. In practice you either learn to cross-check with `list_scenes` or you act on the wrong scene.

kitsupanic:
## Summary

Fix editor_status to report the active editor tab's scene instead of always reporting the game scene.

## Motivation & Context

editor_status was reporting ActiveScene:"Scene" and ActiveScenePath:null regardless of which editor tab was focused. Meanwhile, list_scenes returned the correct scene for the focused tab. This inconsistency broke MCP tooling that relies on editor_status to know which scene is being edited.

The root cause: editor_status read Game.ActiveScene, which is the game scene handle, not the editor tab. The line below it already read the correct source — SceneHasUnsavedChanges comes from SceneEditorSession.Active — and the same resolution idiom appears in SceneTools.ResolveScene's empty-string case.

Before this fix, a single editor_status response could report one scene's name alongside a different scene's dirty flag, since both properties came from different sources.

Fixes: #11639

## Implementation Details

Changed line 146 in TopLevelTools.cs from:
```
var scene = Game.ActiveScene;
```
to:
```
var scene = SceneEditorSession.Active?.Scene ?? Game.ActiveScene;
```

This reads the active editor tab's scene (SceneEditorSession.Active?.Scene) with Game.ActiveScene as fallback — the same source of truth used by SceneHasUnsavedChanges on the next line and SceneTools.ResolveScene elsewhere. No schema or description changes needed; the existing tool description ("the active scene and whether it has unsaved changes") now matches the actual behaviour.

Verification so far: compile-verified (Release, TreatWarningsAsErrors=true, 0 warnings). Behavioural verification over MCP needs an editor restarted against the rebuilt DLL — I'll post the results as a comment here once done (repro steps are in #11639).

None — internal fix, no visual change.

## Checklist

  - Regression test added in Sandbox.Test.Engine: fails on master (ActiveScene comes back null headless), passes with this fix.

### DIFF
--- a/engine/Sandbox.Tools/Mcp/TopLevelTools.cs
+++ b/engine/Sandbox.Tools/Mcp/TopLevelTools.cs
@@ -143,7 +143,7 @@ internal static class TopLevelTools
 	[Description( "Get the current state of the editor - engine version, which project is open, the active scene and whether it has unsaved changes, whether play mode is running or paused, how many tools are registered, the directory paths that matter (logs, project code and assets), and the state of the code compiler. Check IsCompiling/LastCompileSucceeded here after editing code instead of relying on read_console - a successful compile can scroll out of view or be mistaken for the last failure you saw." )]
 	public static EditorStatus GetEditorStatus()
 	{
-		var scene = Game.ActiveScene;
+		var scene = SceneEditorSession.Active?.Scene ?? Game.ActiveScene;
 		var compileGroup = Project.CompileGroup;
 		var buildResult = compileGroup?.BuildResult;
 
--- a/engine/Tests/Sandbox.Test.Engine/Mcp/EditorStatus.cs
+++ b/engine/Tests/Sandbox.Test.Engine/Mcp/EditorStatus.cs
@@ -1,0 +1,116 @@
+using Editor;
+using Editor.Mcp;
+using SceneTests;
+using System;
+
+namespace McpTests;
+
+/// <summary>
+/// <para>
+/// editor_status reports the scene the user is looking at, which is the active editor
+/// session's scene - not the game scene. Reading Game.ActiveScene instead made the tool
+/// report the wrong scene (or none at all) whenever the two differed.
+/// </para>
+/// <para>
+/// Reproduces <see href="https://github.com/Facepunch/sbox-public/issues/11640">Facepunch/sbox-public#11640</see>.
+/// </para>
+/// </summary>
+[TestClass]
+[DoNotParallelize]
+public class EditorStatusTest : SceneTest
+{
+	/// <summary>
+	/// <see cref="TopLevelTools.GetEditorStatus"/> reports FileSystem.Root, which this tier
+	/// never initializes - Bootstrap is what normally does it, and Bootstrap never runs here.
+	/// Point it at the game folder (already the working directory, see the assembly fixture)
+	/// with the base folder mounts skipped, since nothing here reads content.
+	/// </summary>
+	[ClassInitialize]
+	public static void InitializeFilesystem( TestContext context )
+	{
+		if ( EngineFileSystem.Root is null )
+		{
+			EngineFileSystem.Initialize( Environment.CurrentDirectory, skipBaseFolderInit: true );
+		}
+	}
+
+	/// <summary>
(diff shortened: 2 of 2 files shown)


kitsupanic:
Behavioural verification done, source build with this PR applied:

- Idle with three tabs open, editor_status names the focused tab and its path, matching list_scenes' IsActive row exactly (previously "Scene"/null).
- Four tab switches, four correct reports — ActiveScene/ActiveScenePath followed every time, list_scenes agreeing throughout.
- During play it reports via the active game session: Name populated (the played scene), ActiveScenePath null because the game session genuinely has no resource path — list_scenes reports the identical pair for its Game row, so the two tools are now coherent rather than the name being empty.
- After play_stop it returns to naming the active tab with a non-null path (previously this is where it went null).
- SceneHasUnsavedChanges now describes the same scene the name refers to (all-clean session, so weakly exercised).
- Zero console errors across the run.

kitsupanic:
Pushed a regression test for this (Sandbox.Test.Engine): a headless GameEditorSession is made active and the test asserts GetEditorStatus reports its scene's name. Red/green verified — on master the two discriminating asserts fail with ActiveScene null; with this fix all three pass. Full tier run with the test present: 1763/1763 (TestCategory!=LiveBackend, same filter CI uses).

Note for review: GetEditorStatus dereferences FileSystem.Root unconditionally, which is never initialized in Sandbox.Test.Engine, so the test's ClassInitialize calls EngineFileSystem.Initialize guarded on Root being null — commented in the file. If there's a preferred way to arrange that, happy to change it.

DevulTj:
Don't include the tests with this, it's super unnecessary. If it's a one-line fix, PR the one-line fix.

TheLime1:
@DevulTj closing PR for following https://github.com/Facepunch/sbox-public/blob/master/CONTRIBUTING.md
Very good!

PR #11647, opened 2026-08-12
[Mesh Component] Fix blob lifetime during network object creation

SamFadi-dev:
# Pull Request

Thanks for contributing to **s&box** 💖  

---

## Summary

Fixes runtime-networked `MeshComponent` objects missing their polygon mesh on clients that were already connected when the object spawned.

The mesh was displayed correctly for late joiners receiving the initial scene snapshot, but not for clients receiving a live `ObjectCreateMsg` or `ObjectCreateBatchMsg`.

## Motivation & Context

`PolygonMesh` data is serialized through `BlobDataSerializer`.

During live network object creation, component property deserialization is deferred until `CallbackBatch` is disposed. However, the blob context was being disposed before the callback batch flushed.

Consequently, `PostDeserialize` could no longer resolve the polygon mesh blob. The resulting `MeshComponent` existed on the client, but its mesh data was missing and no render mesh could be built.

Reproduction steps:

1. Start a multiplayer session with a host and a connected client.
2. Runtime-spawn and network a prefab containing `MeshComponent`.
3. Observe that the mesh is missing on the already-connected client.
4. Reconnect the client after the object has spawned.
5. Observe that the mesh is displayed correctly through the initial scene snapshot.

## Implementation Details

For `ObjectCreateBatchMsg`, a shared `BlobDataSerializer` context now remains
active for the entire callback batch. Each message blob is loaded only after
the message has passed the existing ownership and duplicate-object
validations.

For an individual `ObjectCreateMsg`, the blob and callback scopes were
reordered so the callback batch is disposed and flushed before the blob
context is disposed.

This preserves the existing validation order and matches the blob lifetime
already used by the initial scene snapshot path.

### Testing

Manually verified using a locally built engine:

- A client connected before the network spawn now receives a valid polygon mesh.
- The `MeshComponent` renders immediately after the live spawn.
- Late-joining clients continue to receive and render the mesh correctly.

No automated regression test has been added yet because reproducing this issue
requires the live multiplayer object-creation and deferred deserialization path.

### Before

Client connected before the runtime network spawn. The networked
`MeshComponent` objects exist, but their polygon geometry is not rendered.

Capture d&#39;écran 2026-08-12 022904

### After

Same scenario with the patched engine. The runtime-networked
`MeshComponent` geometry renders correctly on the already-connected client.

Capture d&#39;écran 2026-08-12 021336

## Checklist

### DIFF
--- a/engine/Sandbox.Engine/Scene/Networking/SceneNetworkSystem.cs
+++ b/engine/Sandbox.Engine/Scene/Networking/SceneNetworkSystem.cs
@@ -1116,6 +1116,8 @@ public partial class SceneNetworkSystem : GameNetworkSystem
 		if ( source is not null && !source.CanSpawnObjects )
 			return;
 
+		using var blobs = BlobDataSerializer.LoadFromMemory( [] );
+
 		using ( CallbackBatch.Batch() )
 		{
 			foreach ( var msg in message.CreateMsgs )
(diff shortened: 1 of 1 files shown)


SamFadi-dev:
issue still here

PR #11684, opened 2026-08-20
EnvmapProbe frame spreading (Performance improvements for realtime probes)

K3rhos:
# Pull Request

## Summary

A realtime "EnvmapProbe" component renders all 6 cubemap faces in a single frame. Thats 6 faces rendering in one spike, if like me you're using probes in Everyframe mode, you do want this to avoid killing your performance!

This PR add a "FrameSpreading" property that deals with those 6 faces refresh across N frames instead.
Default value is 1 so the original behaviour is kept.

## Motivation & Context

I'm using realtime reflection probes as a trickery to get clean and perfect reflection in my scene, I'm making an open world and with a time of day, so baked probes isn't even something to consider, it's just easier to have one giant probe in realtime mode attached to the player/the vehicle, and getting really believable reflections. But currently with no spreads the performance impact was really high, which is now miles better with this PR ! (The only cost being a bit of lagg behind of the probe refresh, but almost imperceptible in most cases)

| FrameSpreading | 720p | 1440p |
|---|---|---|
| 1 (default) | 120 fps | 103 fps |
| 6 | 460 fps | 256 fps |
| 12 | 605 fps | 301 fps |
| 24 | 740 fps | 335 fps |

(For a 1024x1024 realtime probe (the worst case scenario))

~6.8 ms saved at both resolutions.

## Implementation Details

The function `RenderToCubeTexture` already took a face index, so no engine side changes were needed.

GGX filtering runs only after the 6th face. It reads all 6 to build the mip chain, so filtering earlier would filter a mostly stale cubemap, and filtering per face would pay for the expensive half 6 times over.

## Screenshots / Videos

Capture d&#39;écran 2026-08-20 042346
Capture d&#39;écran 2026-08-20 042331
Capture d&#39;écran 2026-08-20 042316
Capture d&#39;écran 2026-08-20 042256

## Checklist

### DIFF
--- a/engine/Sandbox.Engine/Scene/Components/Light/EnvmapProbe.cs
+++ b/engine/Sandbox.Engine/Scene/Components/Light/EnvmapProbe.cs
@@ -65,11 +65,15 @@ public sealed partial class EnvmapProbe : Component, Component.ExecuteInEditor, 
 		TimeInterval,
 	}
 
+	const int NumCubemapFaces = 6;
+
 	SceneCubemap _sceneObject;
 	Texture _dynamicTexture;
 	int _bouncesLeft;
 	int _queuedFrames;
 	float _queuedTime;
+	int _spreadFace;
+	int _spreadIdleFrames;
 
 	public bool Dirty;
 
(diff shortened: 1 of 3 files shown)


K3rhos:
NOTE: I would not consider this a proper solution however, if this PR does get accepted, we should still consider this a workaround more than a real fix, bcs if I recon correctly I remember @sampavlovic (maybe it was someone else) mentioning we should use something called "vulkan multi view" to optimize envmap probe cubemap rendering, but I could be wrong!

PR #11735, opened 2026-08-30
Copy core files using whitelist for standalone export

boxrocket6803:
## Summary
Standalone export previously used a fairly strict whitelist for core, but when base addon was merged into core extra code was added to ensure all base addon files were copied, at the cost of also copying junk files from what used to be core. This instead adds the important files formerly in base addon to the core whitelist and uses only the whitelist to queue files.

## Motivation & Context
I like a clean directory

## Implementation Details

The new whitelist is
```cs
[
	// Shaders
	"shaders/**/*.shader_c",

	// Dev textures, models (contains error assets etc.)
	"textures/dev/**/*.vtex_c",
	"models/dev/**/*.vmdl_c",
	"models/dev/**/*.vmat_c",
	"models/dev/**/*.vtex_c",
	"textures/**/*.vtex_c",
	"debug/*",

	// Default materials
	"materials/**/*.vmat_c",
	"materials/**/*.vtex_c",
	"dev/helper/**/*.vmat_c",
	"dev/helper/**/*.vtex_c",
	"dev/vgui/**/*.vmat_c",
	"dev/*.vtex_c",

	// Splash screen
	"materials/startup_background.vtex_c",

	// Interface
	"fonts/*.ttf",
	"styles/**/*",
	"ui/**/*",

	// Config files
	"cfg/*",
];
```
Which is notably a bit stricter than it was before, there are a few things previously in base now in core that seem generally useless (9mm.ammo_c for example) that I can't imagine standalone games would ever need and would in theory be included via the package references if they were to be used. I've also omitted decals and surfaces, as they reference textures not present in standalone builds (and should also get automatically included if referenced).

## Checklist

### DIFF
--- a/engine/Sandbox.Tools/Utility/Standalone/StandaloneExporter.Files.cs
+++ b/engine/Sandbox.Tools/Utility/Standalone/StandaloneExporter.Files.cs
@@ -34,34 +34,35 @@ partial class StandaloneExporter
 	}
 
 	private static string[] CoreWhitelist = [
-		// Error particle
-		"particles/error/error.vtex_c",
-		"particles/error_particle.vtex_c",
-		
 		// Shaders
-		"shaders/*.shader_c",
+		"shaders/**/*.shader_c",
 		
-		// Dev textures, materials, models (contains error assets etc.)
+		// Dev textures, models (contains error assets etc.)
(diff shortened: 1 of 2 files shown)


handsomematt:
kind of turned on by pack files at this point

PR #11784, opened 2026-09-06
Fix Linux editor multiplayer launch, lobby creation, and joining

aidencurtis:
# Pull Request

## Summary

Fix Linux editor multiplayer testing: launch the native client and dedicated-server executables, create a valid Steam lobby when playing a scene, and let local clients complete the join handshake.

## Motivation & Context

On Linux, **Join via new instance** fails with `System.ComponentModel.Win32Exception (2)` because it tries to start `sbox.exe`, but the installed executable is `sbox`. The dedicated-server action similarly hardcodes `sbox-server.exe`. Removing `.exe` alone is not enough: Unix process lookup does not search the configured working directory for a bare executable name.

After fixing executable startup, testing exposed two more blockers:

- Steam callbacks used Windows' 8-byte struct packing on Linux. The managed lobby-creation result contained a corrupted lobby ID, so hosting failed with "Didn't enter lobby in a reasonable time!" before opening the local TCP listener.
- During client game-code loading, hotload traversed the TCP transport's live queues and cancellation registrations and attempted to rewrite internal .NET socket callbacks. The client then stalled in the join handshake at `MountVPKs`.

Fixes the Linux editor launch, lobby-creation, and local joining failures described above; no linked issue.

## Implementation Details

- In `ViewportTools.Network.cs`, use `Path.Combine(Environment.CurrentDirectory, ...)` for executable paths. Keep `sbox.exe` and `sbox-server.exe` on Windows; use `sbox` and `sbox-server` on other platforms, matching the existing engine convention.
- Define `STEAMWORKS_POSIX` for non-Windows builds and use 4-byte Steam callback packing there. Windows retains 8-byte packing.
- Mark the TCP transport's runtime-owned byte queues, client, and cancellation sources with `[SkipHotload]`. Game message handlers remain eligible for hotload.
- Add regression tests for decoding the native lobby-created callback's Steam ID and for the lobby-entered callback's fields and size.

Working directories and all join, instance ID, window, project, and custom command-line arguments remain unchanged. No game-project or scene changes are required.

### Verification

On Linux:

- Played a scene with `NetworkHelper.StartServer` enabled and confirmed that it created a Steam lobby and opened the local TCP listener.
- Launched two separate native clients against that host. Both completed the handshake, triggered the joined-game callback, and remained `Connected` with network traffic flowing. Host status showed three connected participants.
- Both `SteamTests.LobbyCallbackLayout` regression tests passed.

Earlier startup smoke checks also passed: the native client with `-joinlocal +instanceid 2 -sw -720 +quit` initialized Steam and Vulkan and exited with code 0; the native dedicated server with `+quit` reached its welcome console and exited with code 0. The dedicated server logged mount-loading and missing-shader errors during that startup check.

Dedicated-server joining, remote-machine joining, Windows, and macOS were not runtime-tested. The full test suite was not run.

Not applicable: no visual changes. Runtime verification is described above.

## Checklist

- [x] Public APIs are documented (if applicable): no public API changes
- [x] Unit tests added where applicable and all passing: both added Steam callback tests pass; full suite not run

### DIFF
--- a/engine/Sandbox.Engine/Platform/Steam/Utility/Platform.cs
+++ b/engine/Sandbox.Engine/Platform/Steam/Utility/Platform.cs
@@ -10,7 +10,11 @@ namespace Steamworks
 {
 	internal static class Platform
 	{
+#if STEAMWORKS_POSIX
+		internal const int StructPlatformPackSize = 4;
+#else
 		internal const int StructPlatformPackSize = 8;
+#endif
 		internal const string LibraryName = "steam_api64";
 
 		internal const CallingConvention CC = CallingConvention.Cdecl;
--- a/engine/Sandbox.Engine/Sandbox.Engine.csproj
+++ b/engine/Sandbox.Engine/Sandbox.Engine.csproj
@@ -33,6 +33,10 @@
 		<DefineConstants>WIN;$(DefineConstants)</DefineConstants>
 	</PropertyGroup>
 
+	<PropertyGroup Condition="'$(OS)' != 'Windows_NT'">
+		<DefineConstants>$(DefineConstants);STEAMWORKS_POSIX</DefineConstants>
+	</PropertyGroup>
+
 	<ItemGroup>
 		<PackageReference Include="Facepunch.ActionGraphs" Version="1.9.27" />
 		<PackageReference Include="Fleck" Version="1.2.0" />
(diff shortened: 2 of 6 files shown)


handsomematt:
image


idk

PR #11798, opened 2026-09-07
Expose input rebinding to gamemodes and standalone games

bub-bl:
# Pull Request

---

## Summary

Exposes the engine's input rebinding system to public gamemode code. Currently the whole bind system (`InputBinds`, `BindCollection`) is `internal`, and the only way to remap keys is through `MenuUtility.Input` in the `Sandbox.Menu` assembly — which a gamemode doesn't reference. So gamemodes have **no way** to read or remap their binds programmatically.

This adds a small, public, read/write API on the existing `Sandbox.Input` class so gamemodes can build in-game keybinding UIs (or rebind at runtime) without reaching into engine internals.

## Motivation & Context

Gamemodes need to offer configurable controls, but today they can only:
- rely on `Game.ShowBinds()` (a modal owned by the engine), or
- read binds (read-only `Input.Down/Pressed/Released`, `Input.GetActions`, `Input.GetButtonOrigin`).

There is no supported way to **write** a bind from a gamemode. This PR fills that gap.

This is equally important for **standalone games**. A published standalone build runs the game code directly rather than through the s&box menu, so `MenuUtility.Input` (and the engine's settings screen behind `Game.ShowBinds()`) isn't wired up for it. That leaves standalone titles with no dependency on the menu addon and **no way to remap keys** either — this API provides them the same rebinding capability locally, without shipping the menu.

The rebinding itself is purely local to the client: only action **indices** (bitmasks in `UserCommand.Actions`) are ever sent over the network, so exposing the local mapping introduces no cheating/serialization surface.

## Implementation Details

The new API lives directly on `Sandbox.Input` (namespace `Sandbox`, assembly `Sandbox.Engine`), following the existing pattern where a public `static` wrapper bridges internal engine systems — the same way `ConsoleSystem` wraps the internal `ConVarSystem`.

Because the new methods are in the same assembly as the internals, they access `InputBinds.FindCollection` / `BindCollection` directly with **no** `InternalsVisibleTo` change and **no** widening of the engine's internal surface. The internals remain closed; only a reduced, validated public surface is added:

- `Input.GetBind( action, slot = 0, fallbackToCommon = true )` — resolves game bind → common collection → action default key, mirroring the lookup `Input.OnButton` performs so the returned combo is what the player must actually press.
- `Input.GetAllBinds()` — `Dictionary<string, string[]>` of every action to its two slots.
- `Input.SetBind( action, buttons, slot = 0 )` — writes a bind (takes effect immediately; `Input.OnButton` re-reads the collection on the next press).
- `Input.ResetBinds()` — restores defaults and persists.
- `Input.SaveBinds()` — persists to the JSON config on disk.

**Security / containment considerations**:
- No rebinding affects network data — `UserCommand.Actions` carries only packed action indices (`Connection.Input.cs`), so this is client-local UI/config only.
- The code is precompiled into `Sandbox.Engine`; the user-code sandbox blocks `InternalsVisibleToAttribute` (`Sandbox.Access/Rules/BaseAccess.cs`), so this does not open any additional privilege surface.
- `MenuUtility.Input` (multi-"group" menu API in `Sandbox.Menu`) is intentionally left untouched to avoid a regression; the two APIs coexist.

**Trivia**: `BindEntry` is a struct, not a nullable reference — so `GetAllBinds` reads `FullString` directly (the initial `?.` version failed to compile and was fixed in a follow-up commit).

N/A — public API change; no visuals.

## Checklist


DevulTj:
Not the biggest fan of this. This opens up games to just set all of your binds. I understand the need for this for standalone, but for the platform, it's not ideal.

bub-bl:
> Not the biggest fan of this. This opens up games to just set all of your binds. I understand the need for this for standalone, but for the platform, it's not ideal.

I understand the concern about gamemodes being able to modify the player's binds on the platform, but I think there's also a platform-specific use case here.

On the platform, we can already create our own custom Settings UI, but we currently can't fully manage keybindings through it. Some settings can be handled entirely inside the game, but for keybindings we are forced to open Sandbox's own menus.

For games with a custom UI and UX, this breaks immersion and makes the settings experience inconsistent: the player is suddenly taken out of the game and into Sandbox's UI just to change a keybind.

The goal isn't to give gamemodes unrestricted control over the player's inputs. It's to allow them to manage their own game actions and build a proper custom keybinding UI.

If unrestricted "SetBind" is the concern, I think the API could potentially be restricted to actions belonging to the current gamemode. But I do think this is a genuine platform need, not only a standalone need.

PR #11833, opened 2026-09-11
Blur panel filter layers from a gaussian mip chain

UserDerezzed:
`filter: blur()` and `filter: drop-shadow()` sampled the full-res layer with a 15x15 grid, 225 taps per pixel. The launcher's full-window backdrop blur cost **56 ms of GPU** on an M3 Max, holding it at ~12 fps. At larger radii the taps are also far apart (up to 0.43 sigma), so thin detail comes out as separate copies. You can see this as faint streaks in the launcher backdrop.

This builds a gaussian mip chain for layers that blur, and reads the blur from it.

### Changes
- **New `Graphics.DownsampleMethod.GaussianBlurAlpha`** (`downsample_cs`)
  - Per level: a centred 2x2 box, then the existing 9-tap binomial.
  - Works on all four channels, in linear space, with anything outside the texture transparent.
  - The existing `GaussianBlur` method can't be reused for layers:
    - It forces alpha to 1, breaking drop shadows since they're made from the layer's alpha
    - It samples texel corners, so every level drifts half a texel up-left.
- **`PanelRenderer.Layer.cs`**: generates the chain after the layer is drawn, before the filter quad samples it.
- **`ui/blur.hlsl`**
  - The chain is itself a gaussian, so the shader reads the deepest level whose own blur stays within 90% of the sigma. It covers the rest with 5-9 taps per axis
  - Below the first level it does an exact gaussian on mip 0, with two texels per bilinear fetch.
- **`Panel.Layer.cs`**: sizes the chain with the same rule, capped by the layer's short side (`GenerateMipMaps` stops there).
- **Unblurred reads** (`blur.hlsl`'s zero-sigma path, `ui_borderwrap`) use LOD 0, so a mipped layer drawn scaled down doesn't pick up a blurred mip.

### Results (macOS, M3 Max)
| | Before | After |
|---|---|---|
| Launcher backdrop blur, GPU per frame | 56 ms | 6.5 ms |
| Launcher fps | 12 | 31 |

- **Visually** the result matches/exceeds the original.
  - The launcher backdrop keeps the same softness and brightness.
  - `blur(4px)`, `drop-shadow(6px 6px 10px)` and a filtered child inside a half-opacity parent look identical.
- **Against a true gaussian**, in a 1D model the new path stays within ~9% at every sigma tested. The old grid is exact up to about sigma 2 and drifts far off beyond that.

Screenshots: launcher full, launcher backdrop zoomed, gallery blur and drop-shadow, gallery filter inside opacity.
1-launcher-full
4-gallery-filter-in-opacity
3-gallery-blur-dropshadow
2-launcher-backdrop-zoom

### Testing
- `PanelLayerMips` (new):
  - the chain stops at a thin layer's short side
  - `blur( 8px )` gets exactly 3 mips
  - `blur( 3px )` needs no chain
- UI tests: 1580/1580 pass.
- Manual: Panel Gallery Filters and Blending pages plus the launcher, captured on `master` and this branch.

### Notes
- Only run on macOS so far; Windows and Linux are untested.
- Launcher tops out at 31 fps due to an unrelated issue: https://github.com/Facepunch/mesa-kosmickrisp/pull/2


UserDerezzed:
Rebased onto master after #5785 (Painter). Same change, ported onto the Painter layer path: the mip count for a layer's blur lives on `Painter` now (`Painter.LayerMipCount`), `Painter.Target` and `BeginLayer` create the layer target with that many mips, and the gaussian chain is built right before the composite (`Panel.DrawLayer` → `Painter.GenerateLayerMips`; `LayerScope.Dispose` for public layers). Shader side and the test are unchanged.

PR #11845, opened 2026-09-12
Don't collect garbage on the first ResetEnvironment

UserDerezzed:
## Don't collect garbage on the first ResetEnvironment

`GameInstanceDll.ResetEnvironment` does a full `GC.Collect` + `WaitForPendingFinalizers` so a game's native resources are released before the next one loads. The first time through — `Initialize` and again on the first `LoadGamePackage` — there has been no game, so it's a few hundred milliseconds of startup doing nothing.

### Fix
- `hasResetBefore` flag; the collection runs only once something has actually been reset.

### Results
- Editor: ~330 ms (two resets). Game: measured by the game-startup set ([#11846](https://github.com/Facepunch/sbox-public/pull/11846), [#11847](https://github.com/Facepunch/sbox-public/pull/11847), [#11848](https://github.com/Facepunch/sbox-public/pull/11848), [#11849](https://github.com/Facepunch/sbox-public/pull/11849)) together with the matching TypeLibrary gate ([#11849](https://github.com/Facepunch/sbox-public/pull/11849), stacked on this branch): game-instance init **696 → ~270 ms**; the two interact (skipping the TypeLibrary rebuild alone gains ~90 ms because the forced GC then grows).

### Testing
- Editor opens; game loads a package after the menu (second reset collects as before).

### Part of a set: shortening editor startup

This PR is one of five, each independent, that together cut the time from launching `sbox-dev -project` to the editor being up. All came from tracing the editor boot on an M3 Max / macOS 27; the rows below are the main-thread cost each one removes. Exec → "Bootstrap Init Done", warm caches: **21–26 s → 12.5–16 s (~1.7×)**. Together with [mesa-kosmickrisp#5](https://github.com/Facepunch/mesa-kosmickrisp/pull/5) (driver dlopen: `SourceEnginePreInit` 1.4 s → 0.3 s in the editor) and the launcher set ([#11837](https://github.com/Facepunch/sbox-public/pull/11837)–[#11840](https://github.com/Facepunch/sbox-public/pull/11840)).

| PR | Main-thread cost | Before (ms) | After (ms) | Saved (ms) |
|---|---|---|---|---|
| [Watch folders with FSEvents on macOS](https://github.com/Facepunch/sbox-public/pull/11841) | 49 × `FileSystemWatcher` start | 5,160 | ~0 | ~5,100 |
[...]

PR #11849, opened 2026-09-12
Split ResetEnvironment into teardown and setup; Initialize only sets up

UserDerezzed:
## Split ResetEnvironment into teardown and setup; Initialize only sets up

Reworked after review: no flag. `GameInstanceDll.Initialize` called `ResetEnvironment`, which is written for the moment after a game: it recreates the package loader, drops the enroller, reloads the fonts, rebuilds the game's type library, clears the caches and then runs a full `GC.Collect` + `WaitForPendingFinalizers` so the game's native resources go. At startup none of that has anything to undo - `Bootstrap` built the environment a moment earlier - and just the rebuild and the collection are **~600 ms on the main thread** before the main menu:

```
GameReset - InitTypeLibrary took 265ms
GameReset - GC.Collect+finalizers took ~340ms
IGameMenuDll Bootstrap took 673ms
```

### Fix
- `ResetEnvironment` = `TearDownEnvironment` (everything that undoes a game's use of the environment: loader, enroller, mounted files, fonts, tables, watchers, the type library rebuild, cache clears, the collection) followed by `SetupEnvironment` (everything a clean environment needs before a game: settings and input read fresh, the shared systems reset, an enroller ready to take assemblies).
- `Initialize` runs only `SetupEnvironment`. The three game paths - `CloseGame`, a failed `LoadGamePackageAsync`, and the start of the next load - still call `ResetEnvironment` and behave as before, in the same order.
- Every statement of the old method is in one of the two halves; the only removed line is the `ResetEnvironment()` call in `Initialize`. The teardown steps are all no-ops or pure waste on a fresh environment (empty enroller, nothing mounted, fonts already loaded by the menu, empty tables).

Supersedes #11845 (closed), which gated only the collection behind the same flag; the editor-side saving from that PR is carried here too.

### Results
Game on an M3 Max / macOS 27, medians:
- game-instance init at startup (`IGameInstanceDll.Initialize`) ~700 → ~140 ms
- editor (`sbox-dev -project`): ~330 ms — two resets at boot, each with a full collection (from #11845's trace; the editor set #11841–#11844 links this PR for that row)

### Testing
- Game reaches the main menu; `benchmark +benchmarks facepunch.benchmark` loads a game package and closes it (the reset paths) with no errors in the log.
### Part of a set: shortening game startup

This PR is one of four, each independent, that together cut the time from starting the game to the first menu frame. All came from tracing the game boot on an M3 Max / macOS 27; the rows below are the cost each one removes. Exec → first menu frame, warm caches: **~11.1 s → ~5.9 s (1.9×)**; the first launch after an engine build, which still compiles the menu, ~11.1 s → ~8.4 s. The editor set is [#11841](https://github.com/Facepunch/sbox-public/pull/11841)–[#11845](https://github.com/Facepunch/sbox-public/pull/11845); this PR's `ResetEnvironment` split supersedes #11845.

| PR | Main-thread cost | Before (ms) | After (ms) | Saved (ms) |
|---|---|---|---|---|
| [Cache the menu assembly for the game](https://github.com/Facepunch/sbox-public/pull/11846) | Menu Roslyn compile (`Project.CompileAsync`) | 3,460 | 0 on a cache hit | ~3,450 (every launch but the first after an engine build) |
| [Display modes off the menu critical path](https://github.com/Facepunch/sbox-public/pull/11847) | `SDL_GetFullscreenDisplayModes` in the menu scene load | 980 | 70 | ~910 |
| [Generator: bind only string-literal calls](https://github.com/Facepunch/sbox-public/pull/11848) | Generator `GetSymbolInfo` per invocation, cold compile | 3,460 | 2,600 | ~860 (launches that compile) |
| **Initialize sets the environment up, ResetEnvironment tears down first (this PR)** | `Game.InitTypeLibrary` rebuild + `GC.Collect` on a fresh environment | ~700 | ~140 | ~560 |
| **Total, exec → first menu frame** | | **~11,100** | **~5,900** | **~5,200 (1.9×)** |

Per-row numbers are each PR's own before/after against master, medians of 3–4 runs with `StartupTiming` scopes on the main thread; single phases swung a few hundred ms because the machine was shared with other work, so the rows don't sum exactly to the total.


lolleko (changes requested):
Good idea but this needs a structural refactor not a random flag.

UserDerezzed:
Reworked without the flag. `ResetEnvironment` is now `TearDownEnvironment` (everything that undoes a game's use of the environment: loader, enroller, mounts, fonts, tables, watchers, the type-library rebuild, cache clears, the collection) followed by `SetupEnvironment` (settings and input read fresh, shared systems reset, a fresh enroller). `Initialize` runs only the setup half, since Bootstrap has just built the environment and no game has used it; the three game paths still call `ResetEnvironment` in the same order as before. Every statement of the old method is in one of the two halves; the only removed line is the `ResetEnvironment()` call in `Initialize`. This also covers the GC gate from #11845, so it no longer stacks on it. Rebased onto master after #5785.

PR #11882, opened 2026-09-17
Fix guid-only resource promises never learning their path

PolEpie:
# Pull Request

---

## Summary

Fixes a resource reference that names a path never getting that resource loaded, when a GUID-only
promise for the same GUID already exists. The file is never read, the promise never resolves, and
the reference dangles for the rest of the session.

Surfaces where references are resolved against promises created earlier rather than loaded once
from the asset — scene changes (`Game.ChangeScene`) and network refreshes. Loading the same scene
directly is unaffected. The visible result is a resource property coming back unresolved, or a
`List<GameResource>` coming back empty, with nothing logged.                                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                
## Motivation & Context

`LoadJsonReference` turns `{ "Id": …, "Path": … }` into a `ResourceId` and calls `GetPromise`,
which returns an already-registered object for that GUID as-is:

```cs
var obj = Game.Resources.Get( type, id ) as GameResource;
if ( obj != null ) return obj;   // id.Path is dropped on the floor
 ```

 A load is only ever kicked off by InitPromise, on the new-promise branch, and only when the id
 carries a path:

 ```cs
// a GUID-only promise has no path yet, will only become known by something else finding it
// by GuidIndex and reconciling it (eg. LoadGameResource loading the real file)
if ( !string.IsNullOrEmpty( id.Path ) ) { …; Manifest = AsyncResourceLoader.Load( ResourcePath ); }
 ```

 So a GUID-only reference registers a path-less promise with no load queued, and every later
 reference to that GUID — including ones that do know the path — gets handed that promise back
 without starting one. Nothing loads the file, so nothing reconciles the GUID, so the promise stays
 pending forever.                                                                                                                                                                                                                                                    

## Implementation Details

engine/Sandbox.Engine/Resources/GameResource.cs

 - New StartDeferredLoad( ResourceId ): if this is still a pending promise with no path and the id
   names one, queue AsyncResourceLoader.Load. Nothing else is touched.
 - GetPromise calls it on the object returned by Game.Resources.Get before returning it.

 Deliberately does not assign ResourcePath/ResourceIdLong/ResourceName: those are owned by
 Register, together with the path indexes, and a reference's path is only a hint that may be stale
 (see StalePathPromiseCorrectedOnLoad). The path is still adopted on load, by the existing
 reconciliation. InitPromise is unchanged.

 engine/Tests/Sandbox.Test.Integration/Resources/GuidTests.cs

 - PathReferenceStartsTheLoadForAGuidOnlyPromise: takes a GUID-only promise, asserts it has no
   path and no Manifest, then resolves the same GUID with a path and asserts the same instance
   comes back, a load has been started, and the hinted path has not been adopted yet.

## Checklist


solwllms:
Can you throw up an issue with repro steps/project, please? I don't quite follow what's going on here.

I thought this behaviour was handled already and that we had test coverage for it.

solwllms:
@PolEpie

PolEpie:
Ima trying setup a repro, tbh after GUID update, i got like a scene with component that held a resource, and that wiped everytime i did change scene, but now it doesnt seem to happen anymore (after i got to my private branch with the fix ?)

PolEpie:
@solwllms ok sorry for that but it seem after reload editor it fix by itself and it's hard to do a repro for it, seem due  to me switching to guid update, and then load it again without the update on a older branch

Sorry for this issue !

solwllms:
Happy to hear it haha. Thanks for checking!

